PR #13, opened 2020-03-20
SDKS-293 Support Device Profile Collector Node

witrisna:
Support Device Profile Collector Node

### DIFF
--- a/forgerock-auth-ui/src/main/java/org/forgerock/android/auth/ui/callback/CallbackFragment.java
+++ b/forgerock-auth-ui/src/main/java/org/forgerock/android/auth/ui/callback/CallbackFragment.java
@@ -61,15 +61,26 @@ public abstract class CallbackFragment<T extends Callback> extends Fragment {
         callbackController = null;
     }
 
+    /**
+     * Call when data is collected from the callback.
+     */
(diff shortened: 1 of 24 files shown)


forgerock-auth-ui/src/main/java/org/forgerock/android/auth/ui/callback/CallbackFragment.java
```diff
@@ -70,4 +70,7 @@ public void next() {
         callbackController.next();
     }
 
+    public void cancel(Exception e) {
```
rodrigoareis: Do you think we should include the documentation?

forgerock-auth-ui/src/main/java/org/forgerock/android/auth/ui/callback/ConfirmationCallbackFragment.java
```diff
@@ -45,4 +45,76 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
         return view;
     }
 
+/*
```
rodrigoareis: Are you intentionally leaving this code commented? Should we leave an explanation why this commented out?

PR #14, opened 2020-03-24
Forgerock authenticator

rodrigoareis:


forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/PushParser.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+package org.forgerock.android.authenticator;
+
+import org.forgerock.android.authenticator.exception.URIMappingException;
+import org.forgerock.util.encode.Base64;
+import org.forgerock.util.encode.Base64url;
```
witrisna: These classes are missing

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/PushParser.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.util.Map;
+
+/**
+ * Provides the ability to parse URI scheme into a convenient format
+ * to use with configuring a {@link Push} to receive push notifications.
+ */
+public class PushParser extends UriParser {
```
witrisna: Never use. Do we really need this class?

forgerock-authenticator/src/main/java/org/forgerock/android/auth/DefaultStorageClient.java
```diff
@@ -0,0 +1,236 @@
...
+    private static final String ORG_FORGEROCK_SHARED_PREFERENCES_DATA_NOTIFICATIONS = "org.forgerock.android.authenticator.DATA.NOTIFICATIONS";
+
+    //The SharedPreferences to store the data
+    private SharedPreferences accountData;
+    private SharedPreferences mechanismData;
+    private SharedPreferences notificationData;
+
+    private static final String TAG = DefaultStorageClient.class.getSimpleName();
+
+    public DefaultStorageClient(Context context) {
```
witrisna: I suggest using Builder pattern to build the DefaultStorageClient with SharePreferences interface, the developer can provide a different implementation for storage. Android X provides EncryptedSharePreference if they prefer to use that instead of our implementation.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/DefaultStorageClient.java
```diff
@@ -0,0 +1,236 @@
...
+            Account account = gson.fromJson(entry.getValue().toString(), Account.class);
+            if(account != null)
+                accountList.add(account);
+        }
+
+        return accountList;
+    }
+
+    @Override
+    public boolean removeAccount(Account account) {
```
witrisna: If remove account by id, the input can be just the account id instead of the Account object.

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Account.java
```diff
@@ -0,0 +1,149 @@
...
+    /** Issuer of the account */
+    private final String issuer;
+    /** AccountName, or Username of the account for the issuer */
+    private final String accountName;
+    /** URL of Account's logo image */
+    private final String imageURL;
+    /** HEX Color code in String for Account */
+    private final String backgroundColor;
+    /** An array of Mechanism associated with current Account */
+    private final List<Mechanism> mechanisms;
```
witrisna: This is not used, I like the idea of creating relationship between data model, however, it seem the StorageClient interface did not utilize it. The interface defines store and retrieves independently, they will eventually have data integrity issue, for example, we remove all account but not mechanism, or we remove Push Mechanism but did not remove Notification?

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Account.java
```diff
@@ -0,0 +1,149 @@
...
+    private static final String TAG = Account.class.getSimpleName();
+
+    /**
+     * Creates Account object with given information
+     * @param issuer String value of issuer
+     * @param accountName String value of accountName or username
+     * @param imageURL URL of account's logo image (optional)
+     * @param backgroundColor String HEX code of account's background color (optional)
+     */
+    public Account(String issuer, String accountName, String imageURL, String backgroundColor) {
```
witrisna: Highly recommend using builder pattern with Lombok plugin, it can remove a lot of boilerplate code, like getter, setter, equal, hashcode, and maintenance is easier, when u add an attribute, no implementation is required for getter, equals.....

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Push.java
```diff
@@ -0,0 +1,74 @@
...
+
+import java.util.List;
+
+/**
+ * Represents an instance of a Push authentication mechanism. Associated with an Account.
+ */
+public class Push extends Mechanism {
+
+    /** The endpoint used for registration */
+    public static final String REG_ENDPOINT_KEY = "r";
```
witrisna: Declare private?

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Push.java
```diff
@@ -0,0 +1,74 @@
...
+    /** The challenge to use for the response */
+    public static final String AM_LOAD_BALANCER_COOKIE_KEY = "l";
+
+
+    /** The registration URL for Push mechanism */
+    private String registrationEndpoint;
+    /** The authentication URL for Push mechanism */
+    private String authenticationEndpoint;
+    /** Notifications objects associated with this push mechanism */
+    private final List<Notification> notifications;
```
witrisna: This is not used, I like the idea of creating relationship between data model, however, it seem the StorageClient interface did not utilize it. The interface defines store and retrieves independently, they will eventually have data integrity issue, for example, we remove all account but not mechanism, or we remove Push Mechanism but did not remove Notification?

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Push.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private static final String TAG = Push.class.getSimpleName();
```
witrisna: If not use, remove it.

forgerock-authenticator/build.gradle
```diff
@@ -0,0 +1,215 @@
...
+                }
+            }
+        }
+    }
+}
+
+repositories {
+    maven {
+        credentials {
+            username "$forgerockUser"
```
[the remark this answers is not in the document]
rodrigoareis: Yes, we are using the class org.forgerock.util.encode.Base64 available in Maven.

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/PushParser.java
rodrigoareis: Those class are pulled from gradle:

    implementation 'org.forgerock.commons:json-web-token:3.0.3'

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/PushParser.java
rodrigoareis: Yes, it was design to parse OTP and Push the URLs in their respective models objects. The same approach is used in iOS. But iOS is accepting the URL in the constructor on the model class. I was thinking in a different approach to Android. But I will align it to iOS SDK.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/DefaultStorageClient.java
rodrigoareis: It is a good suggestion, but we were not expecting for them to change the implementation of the DefaultStorageClient class. It implements the StorageClient interface, the one that should be used by them to build their own storage system. But I take a not about it.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/DefaultStorageClient.java
rodrigoareis: It was previously implemented that way. I changed it to align with iOS implementation. I am taking a note to discuss it later.

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Account.java
rodrigoareis: This is just a placeholder, I was thinking about a Lazy initialization for those objects. I believe iOS implemented that way, but it is not decided yet. We will review it and consider this feedback.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/DefaultStorageClient.java
witrisna: With the SharePreference, you can build automation test with standard SharedPreference but not required to run the test as Instrument test.

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Account.java
rodrigoareis: Yes, I am thinking on adopt Lombok, as it is used in fr-auth. I may use Builder to construct those objects as well.

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Push.java
rodrigoareis: This is going to be used by the PushFactory class, so needs to be public.

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Push.java
rodrigoareis: This is just a placeholder, I was thinking about a Lazy initialization for those objects. I believe iOS implemented that way, but it is not decided yet. We will review it and consider this feedback.

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Push.java
rodrigoareis: It is a placeholder, we will definitely use it if accept the URL in the constructor for this class.

forgerock-authenticator/build.gradle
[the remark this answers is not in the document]
witrisna: This is opensource project, how other devs access the repository?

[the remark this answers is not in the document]
rodrigoareis: I think so, I will double check it.

PR #25, opened 2020-04-30
Sdks 354 remove gson library

rodrigoareis:
This PR includes the implementation for the following story:

SDKS-354 [Android][Authenticator SDK] Remove third-party library Gson.

Notes:
- No changes on the unit tests required.


forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Mechanism.java
```diff
@@ -95,6 +98,34 @@ public String getSecret() {
...
+            return null;
+        }
+        try {
+            JSONObject jsonObject = new JSONObject(jsonString);
+            String type = jsonObject.getString("type");
+            if(type.equals(PUSH)) {
+                mechanism = Push.fromJson(jsonString);
+            } else if(type.equals(OATH)) {
+                mechanism = Oath.fromJson(jsonString);
+            }
```
witrisna, line 122: What if not PUSH and OATH? should we throw exception? or expect null?

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Notification.java
```diff
@@ -143,6 +146,68 @@ public final boolean isPending() {
...
+                    .setChallenge(jsonObject.getString("challenge"))
+                    .setAmlbCookie(jsonObject.has("amlbCookie") ? jsonObject.getString("amlbCookie"): null)
+                    .setTimeAdded(jsonObject.has("timeAdded") ? getDate(jsonObject.optLong("timeAdded")) : null)
+                    .setTimeExpired(jsonObject.has("timeExpired") ? getDate(jsonObject.optLong("timeExpired")) : null)
+                    .setTtl(jsonObject.optLong("ttl", -1))
+                    .setApproved(jsonObject.getBoolean("approved"))
+                    .setPending(jsonObject.getBoolean("pending"))
+                    .build();
+        } catch (JSONException e) {
+            return null;
```
witrisna, line 194: Intent to return null if not able to parse it?

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Mechanism.java
rodrigoareis: It can only be PUSH or OATH at this time. It is expected to be null if not able to parse.

forgerock-authenticator/src/main/java/org/forgerock/android/authenticator/Notification.java
rodrigoareis: Yes, it is expected to be null if not able to parse. Updated the method documentation to make it clear.

PR #31, opened 2020-05-20
SDKS-377 Refactor ServerConfig

witrisna:
Refactor ServerConfig, move ServerConfig from core to auth and refactor Config interface


forgerock-core/src/main/java/org/forgerock/android/auth/NetworkConfig.java
```diff
@@ -0,0 +1,85 @@
...
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Manages Network configuration information
+ */
+@EqualsAndHashCode
+@Getter
+class NetworkConfig {
+
+    private String url;
```
rodrigoareis, line 37: Is the url the combination of host, port and realm? I see 'host' as a Network configuration attribute, not sure about 'url' and 'realm'. Do you think makes sense move url and realm to ServerConfig?

Is possible to make the URL nullable? In the Authenticator SDK, for example, we don't need the URL to build a new OkHttpClient. In order to use it today, would be necessary to use the OkHttpClientProvider.getInstance().lookup() every time a new Push authentication mechanism is added, once the registration URL may be different for each mechanism.

witrisna: Make sense, will make the changes.

PR #32, opened 2020-05-27
SDKS-308 Allow developers to customize SDK outbound request, for exam…

witrisna:
Allow developers to customize SDK outbound request, for example customize url to provide query parameters or adding/removing headers

### DIFF
--- a/app/src/main/java/org/forgerock/auth/ForceAuthRequestInterceptor.java
+++ b/app/src/main/java/org/forgerock/auth/ForceAuthRequestInterceptor.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright (c) 2020 ForgeRock. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+package org.forgerock.auth;
+
+import androidx.annotation.NonNull;
+
+import org.forgerock.android.auth.Action;
+import org.forgerock.android.auth.Request;
+import org.forgerock.android.auth.RequestInterceptor;
+
+/**
+ * Sample {@link RequestInterceptor} to add ForceAuth
+ */
+public class ForceAuthRequestInterceptor implements RequestInterceptor {
+
+    @NonNull
+    @Override
+    public Request intercept(@NonNull Request request) {
+        if (((Action)request.tag()).getType().equals("START_AUTHENTICATE")) {
+            return request.newBuilder()
+                    .url(request.url().toString() + "&ForceAuth=true")
+                    .build();
+        }
+        return request;
+    }
+}
--- a/app/src/main/java/org/forgerock/auth/ForceAuthRequestInterceptor.java
+++ b/app/src/main/java/org/forgerock/auth/ForceAuthRequestInterceptor.java
@@ -13,6 +13,8 @@ import org.forgerock.android.auth.Action;
 import org.forgerock.android.auth.Request;
 import org.forgerock.android.auth.RequestInterceptor;
 
+import static org.forgerock.android.auth.Action.START_AUTHENTICATE;
+
 /**
  * Sample {@link RequestInterceptor} to add ForceAuth
  */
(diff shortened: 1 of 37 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/RequestInterceptorFactory.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package org.forgerock.android.auth;
+
+import lombok.Getter;
+
+/**
+ * Factory to manage {@link RequestInterceptor}
+ */
+public class RequestInterceptorFactory {
```
rodrigoareis, line 15: Should this factory be on 'forgerock-core'?

forgerock-core/src/main/java/org/forgerock/android/auth/RequestInterceptor.java
```diff
@@ -0,0 +1,27 @@
...
+public interface RequestInterceptor {
+
+    /**
+     * Intercepts outgoing request from the SDK.
+     *
+     * @param request The original outgoing request
+     * @return The Updated Request
+     */
+    @NonNull
+    Request intercept(@NonNull Request request);
```
rodrigoareis, line 25: I see a difference between the iOS and Android implementation here. iOS has Action as a second parameter. I believe it would also provide a better experience in Android, once the developer does not need to cast the 'tag' attribute from the request to determinate the type of action being intercepted.

app/src/main/java/org/forgerock/auth/ForceAuthRequestInterceptor.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Sample {@link RequestInterceptor} to add ForceAuth
+ */
+public class ForceAuthRequestInterceptor implements RequestInterceptor {
+
+    @NonNull
+    @Override
+    public Request intercept(@NonNull Request request) {
+        if (((Action)request.tag()).getType().equals(START_AUTHENTICATE)) {
```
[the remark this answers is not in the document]
rodrigoareis: See also the comment on "RequestInterceptor.java".

app/src/main/java/org/forgerock/auth/MyCustomDeviceProfile.java
```diff
@@ -0,0 +1,81 @@
...
+    public MyCustomDeviceProfile() {
+    }
+
+    @Keep
+    public MyCustomDeviceProfile(JSONObject jsonObject, int index) {
+        super(jsonObject, index);
+    }
+
+
+        @Override
```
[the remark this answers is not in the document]
witrisna: Removed

app/src/main/java/org/forgerock/auth/InjectHeaderAuthRequestInterceptor.java
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) 2020 ForgeRock. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+package org.forgerock.auth;
```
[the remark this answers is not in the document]
witrisna: I may refactor that later, the whole app is a sample or testing, when it start to get complicate will consider to refactor it.

app/src/main/java/org/forgerock/auth/ForceAuthRequestInterceptor.java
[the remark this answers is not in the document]
witrisna: I would suggest keeping it as is, when the Action gets more attribute e.g payload, the utility method will get complicated.

forgerock-auth/src/main/java/org/forgerock/android/auth/RequestInterceptorFactory.java
witrisna: Moved to core

forgerock-core/src/main/java/org/forgerock/android/auth/RequestInterceptor.java
witrisna: We have discussed this misalignment, seem iOS and JS have no strong opinion for the misalignment, as long as the intent are align. But to make it easier for developer to avoid the cast, I have created FRRequestInterceptor which has the same interface as other platforms. The tag is more generic, we could tag anything instead of bounding it to the Action object. I would like to keep the flexibility.

rodrigoareis: Looks good!

PR #33, opened 2020-05-27
SDKS-383

rodrigoareis:
SDKS-383 [Android] Refactor Authenticator SDK to use OkHttpClientProvider

### DIFF
--- a/forgerock-authenticator/src/main/java/org/forgerock/android/auth/PushResponder.java
+++ b/forgerock-authenticator/src/main/java/org/forgerock/android/auth/PushResponder.java
@@ -163,7 +163,7 @@ class PushResponder {
                 payload.put(DENY_KEY, true);
 
             // Build request
-            OkHttpClient okHttpClient = getOkHttpClient();
+            OkHttpClient okHttpClient = getOkHttpClient(url);
             Request request = buildRequest(url,
                     pushNotification.getAmlbCookie(),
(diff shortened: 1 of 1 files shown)


forgerock-authenticator/src/main/java/org/forgerock/android/auth/PushResponder.java
```diff
@@ -254,18 +254,26 @@ public void onFailure(@NotNull Call call, @NotNull IOException e) {
...
 
-        httpClient = new OkHttpClient.Builder()
-                .connectTimeout(TIMEOUT, SECONDS)
-                .readTimeout(TIMEOUT, SECONDS)
-                .writeTimeout(TIMEOUT, SECONDS)
-                .followRedirects(false)
+        // Extract host from the URL, it's required to build the NetworkConfig and used by
+        // OkHttpClientProvider. There is no need to monitor the the change of hosts considering
+        // the usage of this SDK.
+        String host = url.getProtocol() + "://" + url.getAuthority();
```
witrisna, line 265: You don't need to use the url, I think u can just use url.getAuthority as the host.

rodrigoareis: You are right, I just reused the code from my previous attempt of integration, where the ServerConfig required the 'url' parameter.

PR #41, opened 2020-06-09
SDKS-430 Enable command line test for forgerock-auth and forgerock-core

witrisna:
Run command line to test:
`./gradlew forgerock-core:test`
`./gradlew forgerock-auth:test`


spetrov:
Confirmed that the "core" and "auth" tests run and pass from command line. 
Looks good!

PR #52, opened 2020-07-02
SDKS-495 & SDKS-497 Implement NumberAttributeInputCallback and Boolea…

witrisna:
…nAttributeInputCallback

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/NumberAttributeInputCallback.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/NumberAttributeInputCallback.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright (c) 2019 ForgeRock. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+package org.forgerock.android.auth.callback;
+
+import androidx.annotation.Keep;
+
+import org.json.JSONException;
+import org.json.JSONObject;
+
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+/**
+ * Implements a Callback for collection of a single identity object attribute from a user.
+ */
+@NoArgsConstructor
+@Getter
+public class NumberAttributeInputCallback extends AttributeInputCallback {
+
+    /**
+     * The attribute Value
+     */
+    private Double value;
+
+    /**
+     * Constructor for this Callback.
+     */
+    @Keep
+    public NumberAttributeInputCallback(JSONObject jsonObject, int index) throws JSONException {
+        super(jsonObject, index);
+    }
+
+    @Override
+    protected void setAttribute(String name, Object value) {
+        super.setAttribute(name, value);
+        if ("value".equals(name)) {
+            if (value instanceof Integer) {
+                this.value = ((Integer)value).doubleValue();
+            } else {
+                this.value = (Double) value;
+            }
+        }
+    }
+
+    public void setValue(Double value) {
+        this.value = value;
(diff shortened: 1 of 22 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/callback/NumberAttributeInputCallback.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    @Keep
+    public NumberAttributeInputCallback(JSONObject jsonObject, int index) throws JSONException {
+        super(jsonObject, index);
+    }
+
+    @Override
+    protected void setAttribute(String name, Object value) {
+        super.setAttribute(name, value);
+        if ("value".equals(name)) {
```
rodrigoareis, line 41: May use constant VALUE available in AbstractCallback class?

forgerock-auth/src/test/java/org/forgerock/android/auth/callback/NumberAttributeInputCallbackTest.java
```diff
@@ -0,0 +1,165 @@
...
+                "                {\n" +
+                "                    \"name\": \"IDToken1validateOnly\",\n" +
+                "                    \"value\": false\n" +
+                "                }\n" +
+                "            ]\n" +
+                "        }");
+        NumberAttributeInputCallback callback = new NumberAttributeInputCallback(raw, 0);
+        Assertions.assertThat(callback.getValue()).isEqualTo(123);
+    }
+}
```
rodrigoareis, line 165: Do you think is valid to include a unit test to test a Double as input? I could not find the specification for this callback on the server side, but what is expected if a Float value is provided, should it return null as it is now? Are Double and Integer the only ones supported by this callback on the server?

forgerock-auth/src/test/java/org/forgerock/android/auth/callback/StringAttributeInputCallbackTest.java
```diff
@@ -0,0 +1,115 @@
...
+        Assertions.assertThat(callback.getPrompt()).isEqualTo("Last Name");
+        Assertions.assertThat(callback.isRequired()).isTrue();
+        Assertions.assertThat(callback.getPolicies().getString("name")).isEqualTo("sn");
+        Assertions.assertThat(callback.getFailedPolicies()).isEmpty();
+        Assertions.assertThat(callback.getValidateOnly()).isFalse();
+        Assertions.assertThat(callback.getValue()).isEmpty();
+
+
+    }
+}
```
rodrigoareis, line 115: Do you think is valid to include more unit tests to test the input value, such as non-empty strings, very long strings, special characters, and so on?

forgerock-auth/src/main/java/org/forgerock/android/auth/callback/NumberAttributeInputCallback.java
witrisna: Good catch, updated.

forgerock-auth/src/test/java/org/forgerock/android/auth/callback/NumberAttributeInputCallbackTest.java
witrisna: A test case for double input is added.

forgerock-auth/src/test/java/org/forgerock/android/auth/callback/StringAttributeInputCallbackTest.java
witrisna: The test more focus on parsing the input, there is no special logic on assigning the input json, it just assigned it as is.

PR #55, opened 2020-07-13
Sdks 518

witrisna:


witrisna:
> Am I missing something, or this PR indeed doesn't deal with fr-auth-ui?
> Asking since SDKS-518 AC has this statement: FRUI can handle this property and can generate a realistic UI

I have tested with the FRUI, and able to access those attributes, for the fr-auth-ui, not sure how we want to display the header and description, it is up to the developer how they want to use those attribute.

PR #56, opened 2020-07-14
SDKS-518 Generate a realistic UI for for header and description from …

witrisna:
…Page Node for adaptive UI

### DIFF
--- a/forgerock-auth-ui/src/main/java/org/forgerock/android/auth/ui/callback/AdaptiveCallbackFragment.java
+++ b/forgerock-auth-ui/src/main/java/org/forgerock/android/auth/ui/callback/AdaptiveCallbackFragment.java
@@ -15,6 +15,7 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.Button;
 import android.widget.LinearLayout;
+import android.widget.TextView;
 
 import androidx.annotation.NonNull;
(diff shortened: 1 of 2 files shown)


forgerock-auth-ui/src/main/java/org/forgerock/android/auth/ui/callback/AdaptiveCallbackFragment.java
```diff
@@ -77,6 +79,20 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
         Button nextButton = view.findViewById(R.id.next);
         Button cancelButton = view.findViewById(R.id.cancel);
 
+        TextView header = view.findViewById(R.id.header);
+        if (isEmpty(current.getHeader())) {
+            header.setVisibility(GONE);
```
cerebrl, line 84: What is the effect of `GONE` here?

witrisna: With GONE, the UI element will be removed from the layout, so it won't take any space. There are other options INVISIBLE, VISIBLE. To make it INVISIBLE, the UI element is invisible but still takes some space from the layout.

PR #57, opened 2020-07-16
1. Demo App enhancement

witrisna:
1. Demo App enhancement
2. Align with ForgeRock Theme
3. Allow to choose which tree to invoke
image
image


samples/auth/src/main/java/org/forgerock/auth/TreeDialogFragment.java
```diff
@@ -0,0 +1,73 @@
+/*
+ * Copyright (c) 2019 ForgeRock. All rights reserved.
```
rodrigoareis, line 2: You may need to update the year of the header for new files.

cerebrl:
I know I'm a little late to the party here, but the "START" button seems to be a bit off center (to the right).

PR #61, opened 2020-08-13
SDKS-612 Should not remove Account which is not created by the SDK

witrisna:
When using Android Account manager for SSO, an Account type org.forgerock and Account ForgeRock is created, since only one Account type can be created for an App, the developer may want to create Account for their own App, currently, during user logout with the SDK, the SDK removes all Account under org.forgerock, it should just remove the account created by the SDK.

Few minor enhancement for the PR:

1. Do not remove Account which is not created by SDK
2. Account Name can be provided as a parameter instead of hardcode in the file. (Internal Interface change)
3. Lazy initializing the HttpClient for OAuth2Client

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/AccountSingleSignOnManager.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/AccountSingleSignOnManager.java
@@ -55,7 +55,7 @@ class AccountSingleSignOnManager implements SingleSignOnManager, KeyUpdatedListe
     private Account account;
 
     @Builder
-    public AccountSingleSignOnManager(@NonNull Context context, Encryptor encryptor) throws Exception {
+    AccountSingleSignOnManager(@NonNull Context context, @NonNull String accountName, Encryptor encryptor) throws Exception {
         try {
             this.accountType = getAccountType(context);
         } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/AccountSingleSignOnManager.java
```diff
@@ -55,7 +55,7 @@
     private Account account;
 
     @Builder
-    public AccountSingleSignOnManager(@NonNull Context context, Encryptor encryptor) throws Exception {
+    AccountSingleSignOnManager(@NonNull Context context, @NonNull String accountName, Encryptor encryptor) throws Exception {
```
spetrov, line 58: Isn't that a breaking change? (changing the access modifier and the method signature...) Hopefully customers are not using it in someway...

forgerock-auth/src/main/java/org/forgerock/android/auth/SessionManager.java
```diff
@@ -80,4 +81,23 @@ void close() {
         singleSignOnManager.revoke(null);
     }
 
+    @VisibleForTesting
```
spetrov, line 84: Never heard of this annotation... interesting...

forgerock-auth/src/main/java/org/forgerock/android/auth/AccountSingleSignOnManager.java
witrisna: This is considered internal class, and it is using  builder pattern to build the object, the constructor never been use externally.

PR #70, opened 2020-09-01
SDKS-597 Rollback updated Camera library and fix build issue.

witrisna:


forgerock-authenticator/src/androidTest/java/org/forgerock/android/auth/MockModelBuilder.java
```diff
@@ -37,19 +37,31 @@ public static Account createAccount(String issuer, String accountName, String im
...
-                .setType(oathType)
-                .setAlgorithm(algorithm)
-                .setSecret(secret)
-                .setDigits(digits)
-                .setCounter(counter)
-                .setPeriod(period)
-                .build();
 
-        return oath;
+        switch (OathMechanism.TokenType.TOTP) {
```
rodrigoareis, line 41: should be `switch (OathMechanism.TokenType.TOTP) {`?

spetrov (dismissed):

forgerock-authenticator/src/androidTest/java/org/forgerock/android/auth/MockModelBuilder.java
witrisna: Good catch, updated.

PR #73, opened 2020-10-19
Live e2e test cases

spetrov:
Updated FRAuth e2e test cases to run against live environment. (SDKS-516)


witrisna (approved):
Looks good to me

PR #74, opened 2020-10-19
SDKS-744 Do not call call /endSession when the access token is bound …

witrisna:
Do not call call /endSession when the access token is bound to a session

### DIFF
--- a/forgerock-auth/src/test/java/org/forgerock/android/auth/FRUserMockTest.java
+++ b/forgerock-auth/src/test/java/org/forgerock/android/auth/FRUserMockTest.java
@@ -936,8 +936,6 @@ public class FRUserMockTest extends BaseTest {
                 .addHeader("Content-Type", "application/json")
                 .setBody("{}"));
         enqueue("/sessions_logout.json", HttpURLConnection.HTTP_OK);
-        server.enqueue(new MockResponse()
-                .setResponseCode(HttpURLConnection.HTTP_NO_CONTENT));
 
 
         FRUser.getCurrentUser().logout();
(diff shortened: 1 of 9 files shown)


forgerock-auth/src/test/java/org/forgerock/android/auth/FRUserMockTest.java
```diff
@@ -973,7 +961,7 @@ public void testRequestInterceptor() throws InterruptedException, ExecutionExcep
         Assertions.assertThat(result.get("EXCHANGE_TOKEN").second).isEqualTo(1);
         Assertions.assertThat(result.get("REVOKE_TOKEN").second).isEqualTo(1);
         Assertions.assertThat(result.get("LOGOUT").second).isEqualTo(1);
-        Assertions.assertThat(result.get("END_SESSION").second).isEqualTo(1);
+        //Assertions.assertThat(result.get("END_SESSION").second).isEqualTo(1);
```
rodrigoareis, line 964: Is this comment intentional or left here accidentally ?

rodrigoareis (dismissed):
This commit looks good to me

forgerock-auth/src/test/java/org/forgerock/android/auth/FRUserMockTest.java
witrisna: Good Catch, that is intentional, but the END_SESSION interceptor test is missing. So I have added a test case to test END_SESSION action.

rodrigoareis (approved):
Changes looks good to me

PR #75, opened 2020-10-27
Fixed gradle build from command line

spetrov:
- Fixed build build.gradle in forgerock-auth which was causing build failure from command line.
- Embedded build status label in README.md
- Fixed the 'testShouldSignJWTCorrectly' test to be more reliable (it was failing randomly)


witrisna (approved):
Looks good.

PR #77, opened 2020-10-28
SDKS-649 Refresh Token is not persisted

witrisna:
Refresh Token is not persisted when refresh_token grant is not issuing new Refresh Token


spetrov (approved):
Looks good to me!

PR #78, opened 2020-11-10
org.forgerock.android.auth.FRUser.getAccessToken() clean up tokens in…

witrisna:
… the following conditions: [SDKS-701]

- When Refresh Token Grant Types is used, Server returns invalid_grant (Refresh Token expired), and failed to acquire OAuth2 Tokens with Session Token
- When Refresh Token Grant Types is not used, and failed to acquire OAuth2 Tokens with Session Token


spetrov (approved):
Looks good to me! I tested the fixes and confirmed that they work as expected!

PR #90, opened 2020-12-07
SDKS-808

witrisna:
Persist all attributes for the Cookie and handle HostOnly cookie.

### DIFF
--- a/forgerock-auth/src/androidTest/java/org/forgerock/android/auth/FRDeviceProfileTest.java
+++ b/forgerock-auth/src/androidTest/java/org/forgerock/android/auth/FRDeviceProfileTest.java
@@ -16,6 +16,7 @@ import androidx.test.rule.GrantPermissionRule;
 import org.forgerock.android.auth.collector.*;
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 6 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/CookieMarshaller.java
```diff
@@ -0,0 +1,64 @@
+/*
```
rodrigoareis, line 1: Do you think it would fit on 'core' module?

forgerock-auth/src/androidTest/java/org/forgerock/android/auth/FRDeviceProfileTest.java
```diff
@@ -16,6 +16,7 @@
 import org.forgerock.android.auth.collector.*;
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.junit.Before;
```
rodrigoareis, line 19: Missing year update on the header.

forgerock-auth/src/main/java/org/forgerock/android/auth/SerializableCookie.java
```diff
@@ -0,0 +1,67 @@
+/*
```
rodrigoareis, line 1: Do you think it would fit on 'core' module? Just wondering if it could be reused by other modules in the future

rodrigoareis (commented):
Changes looks good to me, just minor comments to be reviewed.

forgerock-auth/src/main/java/org/forgerock/android/auth/CookieMarshaller.java
witrisna: For now, I would keep in in the auth module, those classes are internal we can easily move it to core when necessary.

forgerock-auth/src/androidTest/java/org/forgerock/android/auth/FRDeviceProfileTest.java
witrisna: Resolved

forgerock-auth/src/main/java/org/forgerock/android/auth/SerializableCookie.java
witrisna: For now, I would keep in in the auth module, those classes are internal we can easily move it to core when necessary.

rodrigoareis (approved):

spetrov (approved):
Looks good to me! 👍

PR #101, opened 2021-04-09
Update input of username

DesarrolloIDPTeco:
Change 
app:passwordToggleEnabled="true"
to
app:passwordToggleEnabled="false"

That input shouldn't have that.


DesarrolloIDPTeco (commented):
Change
app:passwordToggleEnabled="true"
to
app:passwordToggleEnabled="false"

That input shouldn't have that.

PR #106, opened 2021-05-04
SDKS-841 Revisit Rooted Device Detection

witrisna:


forgerock-auth/src/androidTest/java/org/forgerock/android/auth/RootDeviceTest.java
```diff
@@ -34,8 +34,9 @@ public void testDangerousPropertyDetector() {
         RootDetector rootDetector = FRRootDetector.builder()
                 .detector(new DangerousPropertyDetector())
                 .build();
+        int sdkVersion = android.os.Build.VERSION.SDK_INT;
```
rodrigoareis, line 37: Can be `static` and `final` at class level once it's used on multiple tests?

rodrigoareis (approved):
Changes looks good to me

forgerock-auth/src/androidTest/java/org/forgerock/android/auth/RootDeviceTest.java
witrisna: Updated

rodrigoareis (approved):
Looks good to me

PR #109, opened 2021-05-17
SDKS-1039 Align Social Login Custom Handler interface with iOS

witrisna:


spetrov (approved):
Looks good to me! 👍🏻

rodrigoareis (approved):
Changes looks good to me

george-bafaloukas-forgerock (commented):
Looks good! Might be worth adding a test case that you pass a manually selected handler to serve this as an example implementation of the feature

PR #112, opened 2021-05-27
SDKS-1054 Sample APP crashing upon attempt to approve/deny push notif…

rodrigoareis:
This change fix a crash issue on Android devices with API level 26 when trying to load activities that sets ```screenOrientation``` on **AndroidManifest.xml**.

It also includes a new Launch activity to address an issue on app icon not being displayed on some Android devices.


spetrov (approved):
Looks good to me! 👍🏻

PR #116, opened 2021-06-28
Sdks 1060

rodrigoareis:
This change replaces the Firebase Vision library used in the Authenticator sample app by [ML Kit](https://developers.google.com/ml-kit/vision/barcode-scanning/android). It also adds [Zxing from journeyapps](https://github.com/journeyapps/zxing-android-embedded) as fallback in case Google Play Services is not available.


samples/authenticator/src/main/java/org/forgerock/authenticator/sample/camera/CameraScanActivity.java
```diff
@@ -44,40 +50,45 @@
 
     private static final String TAG = CameraScanActivity.class.getSimpleName();
 
+    private PreviewView defaultView;
```
george-bafaloukas-forgerock, line 53: Looks good to me, general note I would prefer not to use generic property names like "defaultView" but that's a minor. Check the commented out code above please as well.

george-bafaloukas-forgerock (approved):
Approved with 2 super minor comments

samples/authenticator/src/main/AndroidManifest.xml
```diff
@@ -76,9 +79,9 @@
             android:configChanges="orientation|screenSize"
             android:label="Notifications"
             tools:ignore="LockedOrientationActivity" />
-        <meta-data
-            android:name="com.google.android.gms.vision.DEPENDENCIES"
-            android:value="barcode" />
+<!--        <meta-data-->
+<!--            android:name="com.google.mlkit.vision.DEPENDENCIES"-->
+<!--            android:value="barcode" />-->
```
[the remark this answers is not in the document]
rodrigoareis: Good catch! Should not be there.

samples/authenticator/src/main/java/org/forgerock/authenticator/sample/camera/CameraScanActivity.java
rodrigoareis: I agree. But the "defaultView" in this case means this will be default view to be used in case Google Play Services is available. Otherwise it will use an alternative view.

spetrov (approved):
Looks good to me!

PR #118, opened 2021-08-18
SDKS 1187

rodrigoareis:
SDKS-1187 Authenticator SDK notification enhancements

**Summary of changes:**
- Make notificationId use date time as plain text
- Remove initialization of PushNotification objects associated with mechanisms inside AuthenticatorManager#getAllAccounts
- Sort list on AuthenticatorManager#getAllNotifications
- Add validation for the primary keys on the build of Models
- Updated unit tests

### DIFF
--- a/forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
+++ b/forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
@@ -100,7 +100,7 @@ class AuthenticatorManager {
         // Retrieve account from StorageClient
         Account account = storageClient.getAccount(accountId);
 
-        // Sets mechanisms and/or notifications associated with this account
+        // Sets mechanisms associated with this account
         initializeAccount(account);
 
         return account;
@@ -171,6 +171,19 @@ class AuthenticatorManager {
         return storageClient.removeNotification(notification);
     }
 
+    PushNotification getNotification(String notificationId) {
+        Logger.debug(TAG, "Retrieving PushNotification with ID '%s' from the StorageClient.", notificationId);
+
+        // Retrieve notification from StorageClient
+        PushNotification notification = storageClient.getNotification(notificationId);
+
+        // Sets mechanism associated with this push notification
+        PushMechanism mechanism = (PushMechanism) getMechanism(notification);
+        notification.setPushMechanism(mechanism);
+
+        return notification;
+    }
+
     void registerForRemoteNotifications(String newDeviceToken) throws AuthenticatorException {
         if(this.deviceToken == null) {
             this.deviceToken = newDeviceToken;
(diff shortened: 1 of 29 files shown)


forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
```diff
@@ -171,6 +171,19 @@ boolean removeNotification(PushNotification notification) {
         return storageClient.removeNotification(notification);
     }
 
+    PushNotification getNotification(String notificationId) {
+        Logger.debug(TAG, "Retrieving PushNotification with ID '%s' from the StorageClient.", notificationId);
+
+        // Retrieve notification from StorageClient
+        PushNotification notification = storageClient.getNotification(notificationId);
```
witrisna, line 178: What if notificationId not found? should return null instead of throw NPE?

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAClient.java
```diff
@@ -211,6 +211,17 @@ public boolean removeMechanism(@NonNull Mechanism mechanism) {
         return this.authenticatorManager.getAllNotifications(mechanism);
     }
 
+    /**
+     * Get the PushNotification object with its id. Identifier of PushNotification object is "<mechanismUUID>-<timeAdded>"
+     * Returns {@code null} if PushNotification could not be found.
```
witrisna, line 216: I think this will throw NPE instead of null if not found.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/OathMechanism.java
```diff
@@ -200,8 +201,15 @@ public T setTimeCreated(Calendar timeCreated) {
         /**
          * Produce the described OathMechanism Token.
          * @return The built Token
+         * @throws MechanismCreationException If an issuer or accountName were not provided.
          */
-        protected OathMechanism build() {
+        protected OathMechanism build() throws MechanismCreationException {
+            if(issuer == null || issuer.isEmpty()) {
```
witrisna, line 207: I think it is worth moving org.forgerock.android.auth.StringUtils from auth to core, and use org.forgerock.android.auth.StringUtils#isEmpty

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
rodrigoareis: It will return null

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAClient.java
rodrigoareis: There was a NullPointException caused by the attempt on set the PushMechanism on the PushNotification object. It was changed to check the null.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/OathMechanism.java
rodrigoareis: I think it's a good idea. But I would recommend handle it in a separated PR.

spetrov (approved):
Looks good to me! 👍🏻

spetrov (commented):
The Jenkins build is choking when trying to build from this branch. 
We need to investigate this failure and fix it before merging it into master.

spetrov (approved):
LGTM! 👍🏻

PR #139, opened 2021-12-15
SDKS-1486 BooleanAttributeInputCallback is not Serializable.

witrisna:
JSONObject defined under AbstractValidatedCallback is not serializable, classes ValidatedUsernameCallback, ValidatePasswordCallback, StringAttributeInputCallback, NumberAttributeInputCallback, BooleanAttributeInoutCallback cannot be serialized when passing the Node object from one Activity to another.

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/AbstractValidatedCallback.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/AbstractValidatedCallback.java
@@ -32,7 +32,7 @@ public abstract class AbstractValidatedCallback extends AbstractCallback {
      *
      * @return validation policies
      */
-    private JSONObject policies;
+    private String policies;
 
     /**
      * Return the list of failed policies for this callback.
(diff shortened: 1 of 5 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/callback/AbstractValidatedCallback.java
```diff
@@ -32,7 +32,7 @@
      *
      * @return validation policies
      */
-    private JSONObject policies;
+    private String policies;
```
rodrigoareis, line 35: Update copyright

rodrigoareis (commented):
Changes looks good to me. We just need to update the header for one file

forgerock-auth/src/main/java/org/forgerock/android/auth/callback/AbstractValidatedCallback.java
witrisna: Updated

rodrigoareis (approved):

PR #140, opened 2022-01-04
SDKS-1368 Provide Build-in Binary Protection to avoid Memory Corruption Attack

witrisna:
SDKS-1368 Provide Build-in Binary Protection to avoid Memory Corruption Attack

Add -fstack-protector-all flag when compile native library for root detection.


spetrov (approved):
LGTM 👍🏻 
I ran the MobSF analysis tool and confirmed that when the sample app is compiled with stack smashing protection, the tool no longer reports the "stack canary" problem in the shared library binary analysis section...

PR #146, opened 2022-03-09
Sdks 1636 Dispose AuthorizationService when no longer required.

witrisna:
SDKS-1636 Dispose AuthorizationService when no longer required.


spetrov (approved):
Changes look good to me!

PR #148, opened 2022-03-23
SDKS-1664 & SDKS-1670

witrisna:
SDKS-1664 - Restore SSO Token when it is out of sync with the SSO Token that bound with the Access Token.
SDKS-1670 - SSO Token should be included in the header instead of request parameter for /authorize endpoint.

This change depends on SDKS-1663


jeyanthanperiyasamy (dismissed):
the changes looks good.

rodrigoareis (dismissed):
Changes looks good to me

spetrov (dismissed):
LGTM 👍🏻

PR #161, opened 2022-05-12
SDKS-1408 Discrepancies between Android and iOS behavior upon registering OATH account (Issuer param)

rodrigoareis:
# JIRA Ticket

[SDKS-1408](https://bugster.forgerock.org/jira/browse/SDKS-1408) Discrepancies between Android and iOS behavior upon registering OATH account (Issuer param)

# Description

While registering a new mechanism, if "issuer" is empty on the query parameter of the URI, but available in the path, the SDK should not throw exception. Instead, it should use the issue in the path.

- [x] Acceptance criteria is met.
- [x] All tasks listed in the user story have been completed.
- [x] Coded to standards.
- [x] Ensure backward compatibility.
- [x] API reference docs is updated.
- [x] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.

### DIFF
--- a/forgerock-authenticator/src/main/java/org/forgerock/android/auth/MechanismParser.java
+++ b/forgerock-authenticator/src/main/java/org/forgerock/android/auth/MechanismParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 ForgeRock. All rights reserved.
+ * Copyright (c) 2020 - 2022 ForgeRock. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -93,6 +93,9 @@ abstract class MechanismParser {
         for (String query : queryParts) {
             String[] split = split(query, "=");
             if (split != null) {
+                if (split[0].equals(ISSUER) && split[1].isEmpty()) {
+                    continue;
+                }
                 r.put(split[0], split[1]);
             }
         }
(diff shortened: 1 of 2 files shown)


spetrov (approved):
Looks good to me! 
I tested the fix using the steps from the problem description in SDKS-1408 and confirmed that the issue no longer occurs... i.e. when "issuer" is empty on the query parameter of the URI but available in the path, the SDK no longer throws an exception.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/MechanismParser.java
```diff
@@ -93,6 +93,9 @@ private Map<String, String> map(URI uri) throws MechanismParsingException {
         for (String query : queryParts) {
             String[] split = split(query, "=");
             if (split != null) {
+                if (split[0].equals(ISSUER) && split[1].isEmpty()) {
```
jeyanthanperiyasamy, line 96: can we also check  if(split != null && split.length > 1) for safer size to avoid index out of bounds

jeyanthanperiyasamy, line 96: can we check this as well  - if (split != null && split.length > 1) . just for safer side to avoid index out of bounds

jeyanthanperiyasamy (changes requested):
minor

jeyanthanperiyasamy (approved):

PR #175, opened 2022-06-28
SDKS-1759 Dynamic config

jeyanthanperiyasamy:
# JIRA Ticket

Please link jira ticket here

# Description

- [ ] Coded to standards.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
@@ -37,6 +37,9 @@ public class Config {
     private String redirectUri;
     private String scope;
     private String oAuthUrl;
+    private Long oauthCacheMillis;
+    private Long oauthThreshold;
+    private Long cookieCacheMillis;
 
     //Server
     private String identifier;
@@ -120,6 +123,9 @@ public class Config {
             userinfoEndpoint = context.getString(R.string.forgerock_userinfo_endpoint);
             logoutEndpoint = context.getString(R.string.forgerock_logout_endpoint);
             endSessionEndpoint = context.getString(R.string.forgerock_endsession_endpoint);
+            oauthCacheMillis = Long.valueOf(context.getResources().getInteger(R.integer.forgerock_oauth_cache) * 1000);
+            oauthThreshold = Long.valueOf(context.getResources().getInteger(R.integer.forgerock_oauth_threshold));
+            cookieCacheMillis = Long.valueOf(context.getResources().getInteger(R.integer.forgerock_cookie_cache) * 1000);
             identifier = UUID.randomUUID().toString();
         }
         initialized = true;
@@ -165,6 +171,8 @@ public class Config {
                 .context(context)
                 .sharedPreferences(sharedPreferences)
                 .oAuth2Client(getOAuth2Client())
+                .cacheIntervalMillis(oauthCacheMillis)
+                .threshold(oauthThreshold)
                 .build();
     }
 
@@ -193,7 +201,7 @@ public class Config {
     private CookieJar getCookieJar() {
         if (cookieJar == null) {
             cookieJar = SecureCookieJar.builder()
-                    .context(context).build();
+                    .context(context).cacheIntervalMillis(cookieCacheMillis).build();
         }
         return cookieJar;
     }
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
@@ -111,7 +111,7 @@ public class Config {
     }
 
 
-    // We need to address this in a different story to remove, because all the tests are depends on this.
+    // We need to remove this in future, because the connected tests are depends on this. but the unit tests/integration tests are not depend on this.
     @Deprecated
     public synchronized void init(Context appContext) {
         if (!initialized) {
@@ -123,27 +123,22 @@ public class Config {
 
     public synchronized void init(Context context, FROptions options) {
         this.context = context.getApplicationContext();
-        cookieJar = null;
-
         clientId = options.getOauth().getOauthClientId();
         redirectUri = options.getOauth().getOauthRedirectUri();
         scope = options.getOauth().getOauthScope();
         oauthCacheMillis = options.getOauth().getOauthCacheSeconds() * 1000;
         oauthThreshold = options.getOauth().getOauthThresholdSeconds();
         cookieCacheMillis = options.getOauth().getCookieCacheSeconds() * 1000;
-
+        cookieJar = null; // We cannot initialize default cookie jar here
         oAuthUrl = options.getServer().getOauthUrl();
         url = options.getServer().getUrl();
         realm = options.getServer().getRealm();
         timeout = options.getServer().getTimeout();
         cookieName = options.getServer().getCookieName();
-
         registrationServiceName = options.getService().getRegistrationServiceName();
         authServiceName = options.getService().getAuthServiceName();
-
         pins = options.getSslPinning().getPins();
         buildSteps = options.getSslPinning().getBuildSteps();
-
         authenticateEndpoint = options.getUrlPath().getAuthenticateEndpoint();
         authorizeEndpoint = options.getUrlPath().getAuthorizeEndpoint();
         tokenEndpoint = options.getUrlPath().getTokenEndpoint();
@@ -151,8 +146,15 @@ public class Config {
         userinfoEndpoint = options.getUrlPath().getUserinfoEndpoint();
         logoutEndpoint = options.getUrlPath().getLogoutEndpoint();
         endSessionEndpoint = options.getUrlPath().getEndSessionEndpoint();
-
         identifier = UUID.randomUUID().toString();
+        FRLogger customLogger = options.getLogger().getCustomLogger();
+        if(customLogger != null) {
+            Logger.setCustomLogger(customLogger);
+        }
+        Logger.Level logLevel = options.getLogger().getLogLevel();
+        if(logLevel != null) {
+            Logger.set(logLevel);
+        }
         initialized = true;
     }
 
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
@@ -60,6 +60,10 @@ public class Config {
     private String logoutEndpoint;
     private String endSessionEndpoint;
 
+    //service
+    private String authServiceName;
+    private String registrationServiceName;
+
     //SSO Token Manager
     private Encryptor encryptor;
 
@@ -82,6 +86,10 @@ public class Config {
         this.url = url;
     }
 
+    public String getUrl() {
+        return this.url;
+    }
+
     private Config() {
     }
 
@@ -102,32 +110,49 @@ public class Config {
         this.sharedPreferences = sharedPreferences;
     }
 
-    public synchronized void init(Context context) {
+
+    // We need to address this in a different story to remove, because all the tests are depends on this.
+    @Deprecated
+    public synchronized void init(Context appContext) {
         if (!initialized) {
-            this.context = context.getApplicationContext();
-            clientId = context.getString(R.string.forgerock_oauth_client_id);
-            redirectUri = context.getString(R.string.forgerock_oauth_redirect_uri);
-            scope = context.getString(R.string.forgerock_oauth_scope);
-            oAuthUrl = context.getString(R.string.forgerock_oauth_url);
-            url = context.getString(R.string.forgerock_url);
-            realm = context.getString(R.string.forgerock_realm);
-            timeout = context.getResources().getInteger(R.integer.forgerock_timeout);
-            cookieJar = null; // We cannot initialize default cookie jar here
-            cookieName = context.getString(R.string.forgerock_cookie_name);
-            pins = Arrays.asList(context.getResources().getStringArray(R.array.forgerock_ssl_pinning_public_key_hashes));
-            buildSteps = Collections.emptyList();
-            authenticateEndpoint = context.getString(R.string.forgerock_authenticate_endpoint);
-            authorizeEndpoint = context.getString(R.string.forgerock_authorize_endpoint);
-            tokenEndpoint = context.getString(R.string.forgerock_token_endpoint);
-            revokeEndpoint = context.getString(R.string.forgerock_revoke_endpoint);
-            userinfoEndpoint = context.getString(R.string.forgerock_userinfo_endpoint);
-            logoutEndpoint = context.getString(R.string.forgerock_logout_endpoint);
-            endSessionEndpoint = context.getString(R.string.forgerock_endsession_endpoint);
-            oauthCacheMillis = Long.valueOf(context.getResources().getInteger(R.integer.forgerock_oauth_cache) * 1000);
-            oauthThreshold = Long.valueOf(context.getResources().getInteger(R.integer.forgerock_oauth_threshold));
-            cookieCacheMillis = Long.valueOf(context.getResources().getInteger(R.integer.forgerock_cookie_cache) * 1000);
-            identifier = UUID.randomUUID().toString();
+            FROptions option = ConfigHelper.load(appContext, null);
+            init(appContext, option);
         }
+        initialized = true;
+    }
+
+    public synchronized void init(Context context, FROptions options) {
+        this.context = context.getApplicationContext();
+        cookieJar = null;
+
+        clientId = options.getOauth().getOauthClientId();
+        redirectUri = options.getOauth().getOauthRedirectUri();
+        scope = options.getOauth().getOauthScope();
+        oauthCacheMillis = options.getOauth().getOauthCacheSeconds() * 1000;
+        oauthThreshold = options.getOauth().getOauthThresholdSeconds();
+        cookieCacheMillis = options.getOauth().getCookieCacheSeconds() * 1000;
+
+        oAuthUrl = options.getServer().getOauthUrl();
+        url = options.getServer().getUrl();
+        realm = options.getServer().getRealm();
+        timeout = options.getServer().getTimeout();
+        cookieName = options.getServer().getCookieName();
+
+        registrationServiceName = options.getService().getRegistrationServiceName();
+        authServiceName = options.getService().getAuthServiceName();
+
+        pins = options.getSslPinning().getPins();
+        buildSteps = options.getSslPinning().getBuildSteps();
+
+        authenticateEndpoint = options.getUrlPath().getAuthenticateEndpoint();
+        authorizeEndpoint = options.getUrlPath().getAuthorizeEndpoint();
+        tokenEndpoint = options.getUrlPath().getTokenEndpoint();
+        revokeEndpoint = options.getUrlPath().getRevokeEndpoint();
+        userinfoEndpoint = options.getUrlPath().getUserinfoEndpoint();
+        logoutEndpoint = options.getUrlPath().getLogoutEndpoint();
+        endSessionEndpoint = options.getUrlPath().getEndSessionEndpoint();
+
+        identifier = UUID.randomUUID().toString();
         initialized = true;
     }
 
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
@@ -12,11 +12,10 @@ import android.content.SharedPreferences;
 
 import androidx.annotation.VisibleForTesting;
 
-import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 
+import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.Setter;
 import okhttp3.CookieJar;
@@ -36,13 +35,14 @@ public class Config {
     private String clientId;
     private String redirectUri;
     private String scope;
-    private String oAuthUrl;
     private Long oauthCacheMillis;
     private Long oauthThreshold;
     private Long cookieCacheMillis;
 
     //Server
+    @Getter(value = AccessLevel.NONE)
     private String identifier;
+
     private String url;
     private String realm;
     private int timeout;
@@ -110,15 +110,11 @@ public class Config {
         this.sharedPreferences = sharedPreferences;
     }
 
-
     // We need to remove this in future, because the connected tests are depends on this. but the unit tests/integration tests are not depend on this.
     @Deprecated
     public synchronized void init(Context appContext) {
-        if (!initialized) {
-            FROptions option = ConfigHelper.load(appContext, null);
-            init(appContext, option);
-        }
-        initialized = true;
+        FROptions option = ConfigHelper.load(appContext, null);
+        init(appContext, option);
     }
 
     public synchronized void init(Context context, FROptions options) {
@@ -130,7 +126,6 @@ public class Config {
         oauthThreshold = options.getOauth().getOauthThresholdSeconds();
         cookieCacheMillis = options.getOauth().getCookieCacheSeconds() * 1000;
         cookieJar = null; // We cannot initialize default cookie jar here
-        oAuthUrl = options.getServer().getOauthUrl();
         url = options.getServer().getUrl();
         realm = options.getServer().getRealm();
         timeout = options.getServer().getTimeout();
@@ -155,7 +150,6 @@ public class Config {
         if(logLevel != null) {
             Logger.set(logLevel);
         }
-        initialized = true;
     }
 
     public static Config getInstance() {
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
@@ -148,12 +148,12 @@ class ConfigHelper: ConfigInterface {
                 context.getSharedPreferences(ORG_FORGEROCK_V_1_HOSTS, Context.MODE_PRIVATE)
             return FROptionsBuilder.build {
                 server {
-                    url = sharedPreferences.getString("url", null)
-                    realm = sharedPreferences.getString("realm", null)
-                    cookieName = sharedPreferences.getString("cookieName", null)
+                    url = sharedPreferences.getString("url", null) ?: ""
+                    realm = sharedPreferences.getString("realm", null) ?: ""
+                    cookieName = sharedPreferences.getString("cookieName", null) ?: ""
                 }
                 oauth {
-                    oauthClientId = sharedPreferences.getString("client_id", null)
+                    oauthClientId = sharedPreferences.getString("client_id", null) ?: ""
                 }
                 urlPath {
                     endSessionEndpoint = sharedPreferences.getString("end_session_endpoint", null)
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
@@ -1,0 +1,164 @@
+/*
+ * Copyright (c) 2022 ForgeRock. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+package org.forgerock.android.auth
+
+import android.content.Context
+import android.content.SharedPreferences
+
+interface ConfigInterface {
+    fun loadFromPreference(context: Context): FROptions
+}
+
+class ConfigHelper: ConfigInterface {
+
+    companion object {
+
+        const val ORG_FORGEROCK_V_1_HOSTS = "org.forgerock.v1.HOSTS"
+
+        @JvmStatic
+        fun persist(context: Context, frOptions: FROptions) {
+            val sharedPreferences =
+                context.getSharedPreferences(ORG_FORGEROCK_V_1_HOSTS, Context.MODE_PRIVATE)
+            sharedPreferences.edit().putString("url", frOptions.server.url)
+                .putString("realm", frOptions.server.realm)
+                .putString("cookieName", frOptions.server.cookieName)
+                .putString("client_id", frOptions.oauth.oauthClientId)
+                .putString("revoke_endpoint", frOptions.urlPath.revokeEndpoint)
+                .putString("end_session_endpoint", frOptions.urlPath.endSessionEndpoint).apply()
+        }
+
+        @JvmStatic
+        fun isConfigChanged(context: Context, frOption: FROptions): Boolean {
+            val sharedPreferences =
+                context.getSharedPreferences(ORG_FORGEROCK_V_1_HOSTS, Context.MODE_PRIVATE)
+            sharedPreferences.getString("url", null)?.apply {
+                if(frOption.server.url != this) {
+                    return true
+                }
+            }
+            sharedPreferences.getString("realm", null)?.apply {
+                if(frOption.server.realm != this) {
+                    return true
+                }
+            }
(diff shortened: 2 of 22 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,119 @@
...
+                    sharedPreferences.edit().putString("end_session_endpoint", frOptions.urlPath.endSessionEndpoint).apply()
+                }
+            }
+
+        @JvmStatic
+        fun isConfigChanged(context: Context?, frOption: FROptions): Boolean {
+            context?.let { safeContext ->
+                val sharedPreferences =
+                    safeContext.getSharedPreferences(ORG_FORGEROCK_V_1_HOSTS, Context.MODE_PRIVATE)
+                sharedPreferences.getString("url", null)?.apply {
```
witrisna, line 41: Will return true if url are equal but not realm?

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,123 @@
...
+    companion object {
+
+        const val ORG_FORGEROCK_V_1_HOSTS = "org.forgerock.v1.HOSTS"
+
+        @JvmStatic
+        fun cache(context: Context?, frOptions: FROptions) {
+                context?.let {
+                    val sharedPreferences =
+                        it.getSharedPreferences(ORG_FORGEROCK_V_1_HOSTS, Context.MODE_PRIVATE)
+                    sharedPreferences.edit().putString("url", frOptions.server.url).apply()
```
witrisna, line 28: Continuously update sharePreferences, sharePreferences.edit().putString().putString().apply()

witrisna (changes requested):

witrisna (changes requested):

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -0,0 +1,134 @@
+package org.forgerock.android.auth
+
+import okhttp3.OkHttpClient
+
+data class FROptions(val server: Server, val oauth: OAuth, val service: Service, val urlPath: UrlPath, val sslPinning: SSLPinning) {
+    companion object {
+        // we cannot do a reference check here, because the reference could be different , but all the option value can be same
+        @JvmStatic
+        fun equals(old: FROptions?, new: FROptions?): Boolean {
+            return old?.oauth == new?.oauth
```
witrisna, line 10: Can we still do reference check + compare the value? so if the reference are same, we don't proceed to compare the value.
if (old !== new {
   //compare value 
} else {
   return true;
}

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,123 @@
...
+        fun cache(context: Context?, frOptions: FROptions) {
+            context?.let {
+                val sharedPreferences =
+                    it.getSharedPreferences(ORG_FORGEROCK_V_1_HOSTS, Context.MODE_PRIVATE)
+                sharedPreferences.edit().putString("url", frOptions.server.url)
+                    .putString("realm", frOptions.server.realm)
+                    .putString("cookieName", frOptions.server.cookieName)
+                    .putString("client_id", frOptions.oauth.oauthClientId)
+                    .putString("revoke_endpoint", frOptions.urlPath.revokeEndpoint)
+                    .putString("end_session_endpoint", frOptions.urlPath.endSessionEndpoint)
```
witrisna, line 33: U need to call apply()

witrisna (changes requested):

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,123 @@
...
+}
+
+class ConfigHelper: ConfigInterface {
+
+    companion object {
+
+        const val ORG_FORGEROCK_V_1_HOSTS = "org.forgerock.v1.HOSTS"
+
+        @JvmStatic
+        fun cache(context: Context?, frOptions: FROptions) {
```
witrisna, line 24: Should we rename to persist()? cache seem caching in memory.

jeyanthanperiyasamy: i like persist

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
jeyanthanperiyasamy: Done , sorry stoyan pulled me for something, forgot to commit

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,123 @@
...
+                    logoutEndpoint = frOption?.urlPath?.logoutEndpoint ?: context.getString(R.string.forgerock_logout_endpoint)
+                    endSessionEndpoint = frOption?.urlPath?.endSessionEndpoint ?: context.getString(R.string.forgerock_endsession_endpoint)
+                }
+            }
+            return frOptions
+        }
+    }
+
+    override fun loadFromCache(context: Context?): FROptions {
+        if(context == null) {
```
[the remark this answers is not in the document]
jeyanthanperiyasamy: at that time context cannot be null, that code needs to be removed

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
jeyanthanperiyasamy: done

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        @JvmStatic
+         fun load(context: Context?, frOption: FROptions?): FROptions {
+            if(context == null) {
+                return FROptionsBuilder.build {}
+            }
+            val frOptions = FROptionsBuilder.build {
+                server {
+                    url = frOption?.server?.url ?: context.getString(R.string.forgerock_url)
```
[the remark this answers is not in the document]
jeyanthanperiyasamy: yeah i was trying to avoid the duplicate code like if and else check, we can be more specific if you like?

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
jeyanthanperiyasamy: sure, yeah we can do an extra check

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,165 @@
...
+                    return true
+                }
+            }
+            return false
+        }
+
+        @JvmStatic
+        fun load(context: Context, frOption: FROptions?): FROptions {
+            return frOption?.let {
+                FROptionsBuilder.build {
```
witrisna, line 65: Any reason we cannot just return the input frOption, but need to clone it?

jeyanthanperiyasamy: @witrisna  this is a perfect catch,  i was moving few things, yes we need to return as it is.

witrisna (changes requested):

forgerock-auth/src/test/java/org/forgerock/android/auth/callback/DeviceProfileCollectorCallbackTest.java
```diff
@@ -12,7 +12,9 @@
 import androidx.test.core.app.ApplicationProvider;
 
 import org.forgerock.android.auth.Config;
+import org.forgerock.android.auth.ConfigHelper;
```
rodrigoareis, line 15: Update copyright

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -0,0 +1,164 @@
+/*
+ * Copyright (c) 2022 ForgeRock. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+package org.forgerock.android.auth
```
rodrigoareis, line 7: Do we need doc for this class?

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright (c) 2022 ForgeRock. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+
+package org.forgerock.android.auth
```
rodrigoareis, line 8: Do we need doc for this class?

rodrigoareis (commented):
Overall implementation looks good to me. Left minor comments.

forgerock-auth/src/test/java/org/forgerock/android/auth/FRAuthMockTest.java
```diff
@@ -214,4 +218,51 @@ public void startTest() {
 
     }
 
+    @Test
+    public void startTestWithOptions() {
```
rodrigoareis, line 222: Can we include a test with multiple starts for different configurations?

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -0,0 +1,164 @@
...
+        return FROptions(server, oauth, service, urlPath, sslPinning, logger)
+    }
+
+}
+
+data class Server(val url: String? = "",
+                  val realm: String? = "root",
+                  val timeout: Int = 30,
+                  val cookieName: String? = "iPlanetDirectoryPro",
+                  val oauthUrl: String? = null)
```
witrisna, line 84: I think we can remove OAuthUrl, it is not using anywhere.

spetrov (changes requested):

forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
```diff
@@ -37,6 +37,9 @@ public class Config {
     private String redirectUri;
     private String scope;
     private String oAuthUrl;
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: forgerock_oauth_url is not used at all in the iOS SDK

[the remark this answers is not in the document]
george-bafaloukas-forgerock: https://github.com/ForgeRock/forgerock-ios-sdk/search?q=forgerock_oauth_url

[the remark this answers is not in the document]
witrisna: Yes, we should remove forgerock_oauth_url

forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
```diff
@@ -37,6 +37,9 @@ public class Config {
...
+    private Long oauthCacheMillis;
```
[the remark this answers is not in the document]
witrisna: By default, those values are 0 which means no caching, however, those values are used in the previous version of SDK already, the encrypt and encrypt process are heavy, it may be good to have this option. We may need to discuss this feature alignment between platforms.

forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
```diff
@@ -37,6 +37,9 @@ public class Config {
...
+    private Long oauthThreshold;
+    private Long cookieCacheMillis;
 
     //Server
     private String identifier;
```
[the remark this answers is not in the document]
witrisna: This attribute is for internal, not necessary to expose it, we can remove the getter.

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
jeyanthanperiyasamy: we can add a doc in the high level

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
jeyanthanperiyasamy: yes we should add a doc for this

forgerock-auth/src/test/java/org/forgerock/android/auth/FRAuthMockTest.java
jeyanthanperiyasamy: Sure , we can back fill more tests for configuration

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,119 @@
...
+    companion object {
+
+        const val ORG_FORGEROCK_V_1_HOSTS = "org.forgerock.v1.HOSTS"
+
+        @JvmStatic
+        fun cache(context: Context?, frOptions: FROptions) {
+                context?.let {
+                    val sharedPreferences =
+                        it.getSharedPreferences(ORG_FORGEROCK_V_1_HOSTS, Context.MODE_PRIVATE)
+                    sharedPreferences.edit().putString("url", frOptions.server.url).apply()
```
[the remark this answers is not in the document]
jeyanthanperiyasamy: fixed

forgerock-auth/src/main/java/org/forgerock/android/auth/FRAuth.java
```diff
@@ -36,24 +35,28 @@ public class FRAuth {
 
     //Alias to store Previous Configure Host
     public static final String ORG_FORGEROCK_V_1_HOSTS = "org.forgerock.v1.HOSTS";
```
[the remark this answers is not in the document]
jeyanthanperiyasamy: fixed

forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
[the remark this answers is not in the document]
jeyanthanperiyasamy: done

forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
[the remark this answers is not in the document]
jeyanthanperiyasamy: so we are keeping it ?

forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
```diff
@@ -57,6 +60,10 @@ public class Config {
     private String logoutEndpoint;
     private String endSessionEndpoint;
```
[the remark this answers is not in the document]
jeyanthanperiyasamy: Config class logic is moved to FROptions , then we do we care about accessing this value publicly ? i am trying to understand this . may be we can sync up

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,129 @@
...
+            sharedPreferences.edit().putString("url", frOptions.server.url)
+                .putString("realm", frOptions.server.realm)
+                .putString("cookieName", frOptions.server.cookieName)
+                .putString("client_id", frOptions.oauth.oauthClientId)
+                .putString("revoke_endpoint", frOptions.urlPath.revokeEndpoint)
+                .putString("end_session_endpoint", frOptions.urlPath.endSessionEndpoint).apply()
+        }
+
+        @JvmStatic
+        fun isConfigChanged(context: Context, frOption: FROptions): Boolean {
```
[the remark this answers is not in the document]
jeyanthanperiyasamy: agree

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
jeyanthanperiyasamy: Done

jeyanthanperiyasamy: done

forgerock-auth/src/test/java/org/forgerock/android/auth/FRAuthMockTest.java
jeyanthanperiyasamy: Done

forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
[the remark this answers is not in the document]
jeyanthanperiyasamy: i will create a new story for this . me and stoyan agree with this.

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,178 @@
...
+                if(frOptions.server.cookieName != this) {
+                    return true
+                }
+            }
+            sharedPreferences.getString("scope", null)?.apply {
+                if(frOptions.oauth.oauthScope != this) {
+                    return true
+                }
+            }
+            sharedPreferences.getString("redirect_uri", null)?.apply {
```
george-bafaloukas-forgerock, line 89: Where did we land on this?
In iOS the FROptions checks for the following:

return (lhs.url == rhs.url &&
        lhs.realm == rhs.realm &&
        lhs.cookieName == rhs.cookieName &&
        lhs.oauthClientId == rhs.oauthClientId)

do we need to align? @jeyanthanperiyasamy @witrisna @spetrov

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,178 @@
...
+         *
+         * @param context  The Application Context
+         * @param frOptions Argument to persist the data in shared preference using the context
+         */
+        @JvmStatic
+        fun persist(context: Context, frOptions: FROptions) {
+            val sharedPreferences =
+                context.getSharedPreferences(ORG_FORGEROCK_V_1_HOSTS, Context.MODE_PRIVATE)
+            sharedPreferences.edit().putString("url", frOptions.server.url)
+                .putString("realm", frOptions.server.realm)
```
george-bafaloukas-forgerock, line 43: Strings like "realm", "cookieName" etc are used multiple times. Could we move them to an enum or statics to avoid re-write and copy paste errors?

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -0,0 +1,178 @@
...
+     *
+     * @param context  The Application Context
+     * @return FROptions from stored persistence
+     */
+    override fun loadFromPreference(context: Context): FROptions {
+            val sharedPreferences: SharedPreferences =
+                context.getSharedPreferences(ORG_FORGEROCK_V_1_HOSTS, Context.MODE_PRIVATE)
+            return FROptionsBuilder.build {
+                server {
+                    url = sharedPreferences.getString("url", null)
```
george-bafaloukas-forgerock, line 163: See comment above for the keys

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -0,0 +1,234 @@
...
+data class FROptions(val server: Server,
+                     val oauth: OAuth,
+                     val service: Service,
+                     val urlPath: UrlPath,
+                     val sslPinning: SSLPinning,
+                     val logger: Log) {
+
+    companion object {
+        @JvmStatic
+        fun equals(old: FROptions?, new: FROptions?): Boolean {
```
george-bafaloukas-forgerock, line 23: This comparison function is different from the isConfigChanged from the ConfigHelper.kt

Why do we have 2 different comparisons? I think this is a protocol one, but in general they should be aligned in the things they compare. We might need to align iOS as well here so we treat the FROptions equality the same in both platforms

george-bafaloukas-forgerock: Unless I am missing something, this "equals", could be used to check if we treat an FRoptions object as "changed" or not based on the rules we have defined. Does it have some extra usage, that requires to check all of the values?

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -0,0 +1,234 @@
...
+                  val timeout: Int = 30,
+                  val cookieName: String? = "iPlanetDirectoryPro")
+
+/**
+ * Server builder to build the SDK configuration information specific to server
+ */
+class ServerBuilder {
+    var url: String? = ""
+    var realm: String? = "root"
+    var timeout: Int = 30
```
george-bafaloukas-forgerock, line 130: Default timeout in iOS is 60s

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+}
+
+/**
+ * Data class for the server configurations
+ */
+data class Server(val url: String? = "",
+                  val realm: String? = "root",
+                  val timeout: Int = 30,
```
george-bafaloukas-forgerock, line 121: Default timeout in iOS is 60

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -0,0 +1,234 @@
...
+                 val cookieCacheSeconds: Long = 0)
+
+/**
+ * Oauth builder to build the SDK configuration information specific to oauth
+ */
+class OAuthBuilder {
+    var oauthClientId: String? = ""
+    var oauthRedirectUri: String? = ""
+    var oauthScope: String? = ""
+    var oauthThresholdSeconds: Long = 30
```
george-bafaloukas-forgerock, line 153: Does the oauthThresholdSeconds need a default value?

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -0,0 +1,234 @@
...
+ * Data class for the Service configurations
+ */
+data class Service(val authServiceName: String? = null,
+                   val registrationServiceName: String? = null)
+
+/**
+ * Service builder to build the SDK configuration information specific to services
+ */
+class ServiceBuilder {
+    var authServiceName: String? = null
```
george-bafaloukas-forgerock, line 171: In iOS we provide default values authServiceName = "Login" & registrationServiceName = "Registration"

Those are the default trees in AM

samples/kotlin/src/main/java/com/forgerock/kotlinapp/FRSessionActivity.kt
```diff
@@ -0,0 +1,115 @@
...
+    private var userInfoFragment: UserInfoFragment? = null
+    private var nodeDialog: NodeDialogFragment? = null
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        setContentView(R.layout.activity_main)
+        updateStatus(showLogin = true)
+        loginButton.setOnClickListener {
+            val journeyName = "Login"
```
george-bafaloukas-forgerock, line 36: Could you add some in line comments explaining the functionality of the FRSession.authenticate() call.

I would also suggest pointing to another tree, to make the distinction of calling separate journeys than the ones in the default configuration.

samples/kotlin/src/main/java/com/forgerock/kotlinapp/FRSessionActivity.kt
```diff
@@ -0,0 +1,115 @@
...
+                    status.visibility = View.GONE
+                    token?.let {
+                        launchUserInfoFragment(token)
+                    }
+
+                }
+            }
+
+            override fun onException(e: Exception?) {
+                TODO("Not yet implemented")
```
george-bafaloukas-forgerock, line 73: Maybe instead of the TODO we can just log the error

forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
```diff
@@ -99,28 +110,48 @@ public void setSharedPreferences(SharedPreferences sharedPreferences) {
         this.sharedPreferences = sharedPreferences;
     }
 
-    public synchronized void init(Context context) {
+    // We need to remove this in future, because the connected tests are depends on this. but the unit tests/integration tests are not depend on this.
+    @Deprecated
+    public synchronized void init(Context appContext) {
```
george-bafaloukas-forgerock, line 115: From an architectural perspective I don't think that the Config class should be a singleton anymore. The fact that we need to have a reset() function that generates a new clean object, breaks slightly the whole pattern. I don't think we will create a memory leak or something so it's not urgent, but in my mind this Config should be hosted as instance property in another class ex: FRAuth that would act as the singleton.

Furthermore, now that we support dynamic configurations this class can be initialized multiple times in the app lifecycle which again breaks the singeton pattern.

Lastly, if someone calls the deprecated public synchronized void init(Context appContext) method and the class is alread initialized, the config will not change. I know that this is not the intented use of the API, but both functions (until the deprecated gets removed) should have the same outcome. There is one scenario that this might happen IRL. Someone starts the SDK by passing an FROptions object and then later calls the default init to re-start the SDK from the Srtings.xml configuration.

george-bafaloukas-forgerock (commented):
@jeyanthanperiyasamy @witrisna @spetrov Let me know if you guys want to discuss any of the comments.

I think we need to align with iOS on the "isChanged" logic. Either Android matches the iOS or vice versa I have no problem either way.

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
george-bafaloukas-forgerock: Sounds good. I'll raise a bug ticket for that and do it

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
jeyanthanperiyasamy: good point

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
jeyanthanperiyasamy: yes i will address this

samples/kotlin/src/main/java/com/forgerock/kotlinapp/FRSessionActivity.kt
jeyanthanperiyasamy: sure

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
jeyanthanperiyasamy: Android uses this , we can talk about this
    <string name="forgerock_auth_service" translatable="false">UsernamePassword</string>
    <string name="forgerock_registration_service" translatable="false">Registration</string>

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
jeyanthanperiyasamy: this was aligned with strings.xml value. the idea is we are going to replicate the values in strings.xml to FROptions for all the default values

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
jeyanthanperiyasamy: i would like to keep the same as 30, since its not been changed , i know its not aligned . not sure it will create side effect for the users and for the backward compatibility let's not change this

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
jeyanthanperiyasamy: i would like to keep the same as 30, since its not been changed , i know its not aligned . not sure it will create side effect for the users and for the backward compatibility let's not change this

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
george-bafaloukas-forgerock: The Strings.xml is passing a value, if you delete that key from there that would default to nothing I guess. That's why I am asking, I think those should be optionals or 0 if possible

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
george-bafaloukas-forgerock: AM "Login" tree is a default tree in AM always present. I would suggest using Login, as I am not sure that out of the box ID Cloud instances have the UsernamePassword tree

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
jeyanthanperiyasamy: nope the purpose is different, there are two types of comparision we are doing , this checks for the local cached options and other is comparing the value from sharedpreference , they are built for different reasons. i am gonna change the method name of isConfigChanged to isConfigDifferentFromPersistedValue .

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
jeyanthanperiyasamy: Done

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
jeyanthanperiyasamy: Andy and i talked about this, we are ok with this to be not aligned

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
jeyanthanperiyasamy: Andy and i talked about this, we are ok with this to be not aligned

spetrov (approved):

forgerock-auth/src/main/java/org/forgerock/android/auth/FRAuth.java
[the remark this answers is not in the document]
spetrov: Constants have been moved in `ConfigHelper` class

forgerock-auth/src/main/java/org/forgerock/android/auth/FRAuth.java
```diff
@@ -36,24 +35,28 @@ public class FRAuth {
...
+
     private static boolean started;
 
-    public static void start(Context context) {
-        if (!started) {
+    private static FROptions cachedOptions;
+
+    public static void start(Context context, @Nullable FROptions options) {
```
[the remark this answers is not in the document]
spetrov: This has been addressed by Jey.

PR #180, opened 2022-07-14
Dynamic sdk integration test cleanup

jeyanthanperiyasamy:
# JIRA Ticket

Please link jira ticket here

# Description

- [ ] Coded to standards.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.


forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
```diff
@@ -107,13 +107,22 @@ public void setSharedPreferences(SharedPreferences sharedPreferences) {
         this.sharedPreferences = sharedPreferences;
     }
 
-    // We need to remove this in future, because the connected tests are depends on this. but the unit tests/integration tests are not depend on this.
-    @Deprecated
+    /**
+     * Load all the SDK configuration from the strings.xml
+     *
+     * @param appContext  The Application Context
+     */
```
jeyanthanperiyasamy, line 114: Actually i like to keep this method. this would be the easiest way to load the values from strings.xml. may be some customer may like to use it when they call after **reset()** method

witrisna (dismissed):

forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
```diff
@@ -107,15 +108,15 @@ public void setSharedPreferences(SharedPreferences sharedPreferences) {
         this.sharedPreferences = sharedPreferences;
     }
 
-    // We need to remove this in future, because the connected tests are depends on this. but the unit tests/integration tests are not depend on this.
-    @Deprecated
```
spetrov, line 111: This is a breaking change, right? We need to publish a major release and have a note in the changelog about this potential breaking change... Let's bring this to the team because there might be other things that I'm forgetting...

spetrov (commented):
Looks good to me. But since there is a breaking change, we need to discuss with the team about how to release this...

forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
witrisna: Developer should not call this interface, consider this an internal class. Most of the method under Config are more for testing purpose, except some special use cases.

PR #183, opened 2022-07-25
SDKS - 2031 Bugs fixed when change on client_id to load from persisted options

jeyanthanperiyasamy:
# JIRA Ticket

Please link jira ticket here

# Description

- [ ] Coded to standards.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
@@ -8,34 +8,27 @@
 package org.forgerock.android.auth
 
 import android.content.Context
-import android.content.SharedPreferences
-
-/**
- * Listener to load the options from persistence for the given context
- */
-interface ConfigInterface {
-    fun loadFromPreference(context: Context): FROptions
-}
 
 /**
  * ConfigHelper Provide the helper methods to persist and load the configurations on start of the SDK
  * This class also verify if the configuration changes on the current and previous launch.
  */
-internal class ConfigHelper: ConfigInterface {
+internal class ConfigHelper {
 
     companion object {
 
-         const val realm = "realm"
-         const val url = "url"
-         const val cookieName = "cookieName"
-         const val clientId = "client_id"
-         const val revokeEndpoint = "revoke_endpoint"
-         const val endSessionEndpoint = "end_session_endpoint"
-         const val scope = "scope"
-         const val redirectUri = "redirect_uri"
+        private const val realm = "realm"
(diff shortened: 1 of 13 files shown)


witrisna (approved):

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
```diff
@@ -8,34 +8,27 @@
...
 
-         const val realm = "realm"
-         const val url = "url"
-         const val cookieName = "cookieName"
-         const val clientId = "client_id"
-         const val revokeEndpoint = "revoke_endpoint"
-         const val endSessionEndpoint = "end_session_endpoint"
-         const val scope = "scope"
-         const val redirectUri = "redirect_uri"
+        private  const val realm = "realm"
```
george-bafaloukas-forgerock, line 20: super minor, the `private  const val realm = "realm"` has two spaces before const

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -147,9 +147,9 @@ data class OAuth(val oauthClientId: String? = "",
  * Oauth builder to build the SDK configuration information specific to oauth
  */
 class OAuthBuilder {
-    var oauthClientId: String? = ""
-    var oauthRedirectUri: String? = ""
-    var oauthScope: String? = ""
+    var oauthClientId: String = ""
```
george-bafaloukas-forgerock, line 150: Will the non optional v alus allow the app to be used without setting an OAuth client at all? If they opt to use the FRSession classes, they can run the authentication without running the OAuth flows

jeyanthanperiyasamy: can we talk about this more, trying to understand this

jeyanthanperiyasamy: @george-bafaloukas-forgerock  . Both me and @witrisna  tested this flow for both static and dynamic config, we are able to get the SSO Token for FRsession.authenticate. Thanks for bringing this scenario to test

forgerock-auth/src/main/java/org/forgerock/android/auth/ConfigHelper.kt
jeyanthanperiyasamy: done

witrisna (approved):

PR #222, opened 2023-01-10
SDKS-2091 Remove Keys after the binding process is failed

witrisna:
# JIRA Ticket

SDKS-2091

# Description

SDKS-2091 Remove Keys after the binding process is failed
When the binding process is triggered, it generates new keys and replaces existing keys, after authentication failed, the newly generated keys are left behind in the device. During the process of Device Signing Verification, since there is a generated keys, it triggered the authentication process, but the keys do not match with the server, and make the signing process failed. To resolve this, when authentication failed, it removes the keys, and the user is required to perform rebinding again.

- [ ] Coded to standards.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 ForgeRock. All rights reserved.
+ * Copyright (c) 2022 - 2023 ForgeRock. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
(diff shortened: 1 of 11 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
```diff
@@ -194,41 +194,41 @@ open class DeviceBindingCallback : AbstractCallback, Binding {
...
-                        keyPair.keyAlias,
-                        deviceBindingAuthenticationType)
-                    val jws = deviceAuthenticator.sign(keyPair,
-                        kid,
-                        userId,
-                        challenge,
-                        getExpiration(timeout))
-                    setJws(jws)
-                    setDeviceId(deviceId)
+                    keyPair?.let {
```
jeyanthanperiyasamy, line 208: is the KeyPair is optional ? if not we can remove the check keyPair?.let

forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
```diff
@@ -194,41 +194,41 @@ open class DeviceBindingCallback : AbstractCallback, Binding {
...
         }
 
+        //TODO We may need to delete other keys if we only want to maintain one keys on the device
+        //TODO However, we may want to have Application Pin as fallback, so for now, keep multiple keys.
+        var keyPair: KeyPair? = null
         try {
-            val keyPair: KeyPair
             val status: DeviceBindingStatus
             withTimeout(getDuration(timeout)) {
                 keyPair = deviceAuthenticator.generateKeys(context);
```
jeyanthanperiyasamy, line 203: Remove the colon

jeyanthanperiyasamy (commented):
minor feedback

forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
```diff
@@ -194,41 +194,41 @@ open class DeviceBindingCallback : AbstractCallback, Binding {
             return
         }
 
+        //TODO We may need to delete other keys if we only want to maintain one keys on the device
+        //TODO However, we may want to have Application Pin as fallback, so for now, keep multiple keys.
+        var keyPair: KeyPair? = null
```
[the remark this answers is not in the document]
witrisna: It can be null, when deviceAuthenticator.generateKeys failed with Exception or timeout, the keyPair can be null, and would like to remove both the reference (from SharedPreference) and the actual Key if anything failed.

forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
witrisna: Removed

spetrov (approved):
👍🏻

PR #229, opened 2023-01-23
SDKS-2293 Developer interface to unbind device

witrisna:
# JIRA Ticket

[SDKS-2293](https://bugster.forgerock.org/jira/browse/SDKS-2293)

# Description

SDKS-2293 Developer interface to unbind device

- [ ] Coded to standards.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/FRUserKeys.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/FRUserKeys.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2023 ForgeRock. All rights reserved.
+ *
+ *  This software may be modified and distributed under the terms
+ *  of the MIT license. See the LICENSE file for details.
+ */
+
+package org.forgerock.android.auth
+
+import android.content.Context
+import org.forgerock.android.auth.callback.DeviceBindingCallback
+import org.forgerock.android.auth.devicebind.UserDeviceKeyService
+import org.forgerock.android.auth.devicebind.UserKey
+import org.forgerock.android.auth.devicebind.UserKeyService
+
+/**
+ * Manage [UserKey] that created by the SDK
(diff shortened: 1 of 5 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/FRUserKeys.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Delete the [UserKey]
+     * @param userKey The [UserKey] to be deleted
+     */
+    fun delete(userKey: UserKey) {
+        userKeyService.delete(userKey)
+    }
+
+}
```
vahancouver, line 39: In the ticket it says "Provide interface to delete all keys.". Are you gonna have a separate method for that or developers should get all the keys and delete one by one?

jeyanthanperiyasamy: +1 , i like to have deleteAll()

jeyanthanperiyasamy (commented):
some feedback

forgerock-auth/src/main/java/org/forgerock/android/auth/FRUserKeys.kt
witrisna: I may update the ticket to not delete all keys, in the future, I am thinking of using the same interface to delete keys not only locally but also from the server as well, since user can only delete his/her own key with the session, delete all key make not make sense.

@jeyanthanperiyasamy Yes, has a look at the UserDeviceKeyService, it not only delete the reference, but also the keys, but not sure that is the best class to handle it.

jeyanthanperiyasamy: this point make sense to me...i didnt think of multiple user can also login in same device. lets keep the same interface

spetrov (approved):

PR #230, opened 2023-01-23
SDKS-2282 - fix to devicebind failure in Android 7 using PKCS12

jeyanthanperiyasamy:
# JIRA Ticket

Please link jira ticket here

# Description

- [ ] Coded to standards.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/ApplicationPinDeviceAuthenticator.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/ApplicationPinDeviceAuthenticator.kt
@@ -19,6 +19,7 @@ import org.forgerock.android.auth.devicebind.DeviceBindingErrorStatus.Abort
 import org.forgerock.android.auth.devicebind.DeviceBindingErrorStatus.UnAuthorize
 import org.forgerock.android.auth.devicebind.DeviceBindingErrorStatus.UnRegister
 import java.io.FileNotFoundException
+import java.io.IOException
 import java.io.InputStream
 import java.io.OutputStream
 import java.security.KeyStore
@@ -63,11 +64,9 @@ open class ApplicationPinDeviceAuthenticator(private val pinCollector: PinCollec
     }
 
     override suspend fun authenticate(context: Context): DeviceBindingStatus {
-
         if (!appPinAuthenticator.exists(context)) {
             return UnRegister()
         }
-
         var pin = pinRef.getAndSet(null)
         pin?.let {
             return getPrivateKey(context, it)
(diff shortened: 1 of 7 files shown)


forgerock-auth/src/test/java/org/forgerock/android/auth/devicebind/ApplicationPinDeviceAuthenticatorTest.kt
```diff
@@ -124,7 +124,7 @@ class ApplicationPinDeviceAuthenticatorTest {
         authenticator2.prompt(prompt)
         authenticator2.setKey(cryptoKey)
         val status = authenticator2.authenticate(context)
-        assertThat(status).isEqualTo(UnAuthorize())
```
witrisna, line 127: Is there a way to use UnAuthorize when the Pin is invalid?
Unreg - The device has not been registered
UnAuth - The device is registered, but user provide the wrong pin to unlock.

forgerock-core/src/test/java/org/forgerock/android/auth/AppPinAuthenticatorTest.kt
```diff
@@ -83,13 +89,4 @@ class AppPinAuthenticatorTest {
             Offset.offset(1000))
     }
 
-    @Test
-    fun testExists() {
```
jeyanthanperiyasamy, line 87: we canot test this now, because now its just looking the actual file exist ..and its mocked already

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/ApplicationPinDeviceAuthenticator.kt
```diff
@@ -98,6 +97,8 @@ open class ApplicationPinDeviceAuthenticator(private val pinCollector: PinCollec
             UnRegister()
         } catch (e: UnrecoverableKeyException) {
             UnAuthorize()
```
witrisna, line 99: May be provide code comment here about IOException and UnrecoverableKeyException

forgerock-core/src/test/java/org/forgerock/android/auth/AppPinAuthenticatorTest.kt
witrisna: I think we can implement the exists method for the overridden KeyStoreRepository (The one using byteArrayOutputStream) and check if the byte array has data

jeyanthanperiyasamy: i like that idea , we can do like byteArrayOutputStream.toByteArray().isNotEmpty(). Let me change that

spetrov (approved):
LGTM! 👍🏻

PR #239, opened 2023-02-07
SDKS-2327 Authenticator SDK - jailbreak and biometric policies

rodrigoareis:
# JIRA Ticket

[SDKS-2327](https://bugster.forgerock.org/jira/browse/SDKS-2327) [Android] Authenticator SDK - jailbreak and biometric policies

# Description

This PR introduces the Policy Evaluator, an API used by the SDK to enforce Policy rules, such as Device Tampering Policy. This will work along the changes introduced on [SDKS-2165](https://bugster.forgerock.org/jira/browse/SDKS-2165).

- [ ] Coded to standards.
- [ ] Unit tests are written.
- [ ] e2e tests are written.
- [ ] Change log updated.


forgerock-authenticator/src/main/java/org/forgerock/android/auth/Account.java
```diff
@@ -38,21 +39,33 @@ public class Account extends ModelObject<Account> {
...
-     * @param backgroundColor String HEX code of account's background color (optional)
-     * @param timeAdded Date and Time this Account was stored
+     * @param imageURL           URL of account's logo image (optional)
+     * @param backgroundColor    String HEX code of account's background color (optional)
+     * @param timeAdded          Date and Time this Account was stored
+     * @param policies           Policies used to enforce device security (optional)
+     * @param lockingPolicy      Indicates the policy locking the account (optional)
+     * @param lock               Indicates if the account is locked (optional)
      */
     public Account(String issuer, String displayIssuer, String accountName, String displayAccountName,
```
witrisna, line 65: Do we need to care about backward compatible? This will release as 4.0?

forgerock-authenticator/src/main/java/org/forgerock/android/auth/Account.java
```diff
@@ -182,6 +238,9 @@ public String toJson() {
             jsonObject.put("imageURL", imageURL);
             jsonObject.put("backgroundColor", backgroundColor);
             jsonObject.put("timeAdded", timeAdded != null ? timeAdded.getTimeInMillis() : null);
+            jsonObject.put("policies", policies);
```
witrisna, line 241: I think that will treat policies as String instead of JSONObject.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/Account.java
```diff
@@ -208,6 +267,9 @@ public static Account deserialize(String jsonString) {
                     .setImageURL(jsonObject.has("imageURL") ? jsonObject.getString("imageURL") : null)
                     .setBackgroundColor(jsonObject.has("backgroundColor") ? jsonObject.getString("backgroundColor") : null)
                     .setTimeAdded(jsonObject.has("timeAdded") ? getDate(jsonObject.optLong("timeAdded")) : null)
+                    .setPolicies(jsonObject.has("policies") ? jsonObject.getString("policies") : null)
```
witrisna, line 270: May need to pay attention to this, the underlining may be a JSONObject, but we convert it to String.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
```diff
@@ -128,6 +142,29 @@ boolean updateAccount(Account account) {
         }
     }
 
+    boolean lockAccount(Account account, FRAPolicy policy) throws AccountLockException {
+        if (account.isLocked()) {
+            throw new AccountLockException("Error locking the Account: Account is already locked.");
+        } else if (policy.getName() == null || policy.getName().isEmpty()) {
+            throw new AccountLockException("Error locking the Account: The policy name is required.");
+        } else if (!account.getPolicies().contains(policy.getName())) {
```
witrisna, line 150: Recommend mapping the policy name instead of the whole string.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
```diff
@@ -240,9 +289,69 @@ List<PushNotification> getAllNotifications(@NonNull Mechanism mechanism) {
...
+            public void onException(Exception e) {
+                Logger.error(TAG, "Error creating OATH mechanism in Combined MFA.");
+            }
+        });
+
+        // Register PUSH mechanism passing combined parameters
+        pushFactory.createFromUri(uri, new FRAListener<Mechanism>() {
+            @Override
+            public void onSuccess(Mechanism result) {
+                listener.onSuccess(result);
```
witrisna, line 325: We just want to return one result?

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
```diff
@@ -240,9 +289,69 @@ List<PushNotification> getAllNotifications(@NonNull Mechanism mechanism) {
...
+                Logger.debug(TAG, "PUSH mechanism in Combined MFA successfully created.");
+            }
+
+            @Override
+            public void onException(Exception e) {
+                listener.onSuccess(oathMechanism[0]);
```
witrisna, line 331: Does the caller receive success and exception at the same time?

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
```diff
@@ -240,9 +289,69 @@ List<PushNotification> getAllNotifications(@NonNull Mechanism mechanism) {
...
+        });
+    }
+    
     private void initializeAccount(Account account) {
         if(account != null) {
-            Logger.debug(TAG, "Loading associated data for the Account with ID: %s", account.getId());
+            Logger.debug(TAG, "Evaluating policies for the new Account with ID: %s",
+                    account.getId());
+            boolean policiesCompliance = policyEvaluator.evaluate(context, account);
+            if(!account.isLocked() & !policiesCompliance) {
```
witrisna, line 343: In purpose doing & instead of &&?

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
```diff
@@ -240,9 +289,69 @@ List<PushNotification> getAllNotifications(@NonNull Mechanism mechanism) {
...
+                Logger.debug(TAG, "Lock Account due non-compliance policy with name: %s",
+                        policyEvaluator.getNonCompliancePolicy().getName());
+                account.lock(policyEvaluator.getNonCompliancePolicy());
```
witrisna, line 346: Looks like a bit danger to maintain state in the policyEvaluator, if calling the initializeAccount with a different Account, keeping the nonCompliancePolicy may be an issue.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAClient.java
```diff
@@ -162,8 +183,9 @@ public Account getAccount(@NonNull Mechanism mechanism) {
      * not be found or updated.
      * @param account The Account to update.
      * @return boolean as result of the operation
+     * @throws AccountLockException if account is locked
      */
-    public boolean updateAccount(@NonNull Account account) {
+    public boolean updateAccount(@NonNull Account account) throws AccountLockException {
```
witrisna, line 188: For updateAccount, lockAccount, unlockAccount, the interface return boolean, but at the same time throw AccountLockException, should we just use either one (Prefer AccountLockException), if exception throw, we consider failed, otherwise success. This will make developer easier to handle the error scenario.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAPolicyEvaluator.java
```diff
@@ -0,0 +1,169 @@
...
+     * The asynchronous PolicyEvaluator builder.
+     */
+    public static class FRAPolicyEvaluatorBuilder {
+        private List<FRAPolicy> policies;
+
+        /** Sets a custom list of policies.
+         * @param policies The list with custom policies.
+         * @return this builder
+         */
+        public FRAPolicyEvaluatorBuilder withPolicies(@NonNull List<FRAPolicy> policies) {
```
witrisna, line 67: It will be good if we have withPolicy(FRAPolicy policy), so developer can do 
FRAPolicyEvaluatorBuilder().withPolicy(BiometricPolicy()).withPolicy(DeviceTemperingPolicy()).build()

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAPolicyEvaluator.java
```diff
@@ -0,0 +1,169 @@
...
+    private static final String TAG = FRAPolicyEvaluator.class.getSimpleName();
+
+    public static final List<FRAPolicy> DEFAULT_POLICIES = new ArrayList<>();
+
+    static {
+        DEFAULT_POLICIES.add(new DeviceTamperingPolicy());
+        DEFAULT_POLICIES.add(new BiometricAvailablePolicy());
+    }
+
+    private FRAPolicy nonCompliancePolicy;
```
witrisna, line 41: Maintaining state here may be dangerous, this makes the Evaluator can use only evaluate once.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/MechanismParser.java
```diff
@@ -146,5 +161,31 @@ private static String stripSlash(String s) {
...
+        byte[] bytes = Base64.decode(value, Base64.NO_WRAP + Base64.URL_SAFE);
+        return new String(bytes);
+    }
+
+    protected boolean isBase64Encoded(String value) {
+        try {
+            Base64.decode(value, Base64.DEFAULT);
+            return true;
+        } catch (Exception e) {
+            return false;
```
witrisna, line 186: May be useful to provide debug log.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/policy/FRAPolicy.java
```diff
@@ -0,0 +1,50 @@
...
+     */
+    public abstract String getName();
+
+    /**
+     * Evaluate the policy compliance.
+     *
+     * @param context  The Application Context.
+     * @return {@code true} if the policy comply, {@code false} otherwise
+     */
+    public abstract boolean evaluate(Context context);
```
witrisna, line 48: Something to think of not necessary to change it.
We may want to make the evaluate interface async like evaluate(context, FRListener), for example, if we want to have LocationAwarePolicy, getting the location from device is using callback, just in case we need to perform something and waiting for input.

witrisna (changes requested):

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAPolicyEvaluator.java
witrisna: Similar pattern is using under org.forgerock.android.auth.AuthService#AuthService, but I am using lombok, with @Builder and @Singluar, you can delombok it and see how they generate the inteface.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAPolicyEvaluator.java
rodrigoareis: Good point, but I think we could clean this attribute when starting a new evaluation instead of remove this property. How that sounds?

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAPolicyEvaluator.java
rodrigoareis: I like the idea. Lets see if we can add it easily

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAPolicyEvaluator.java
witrisna: With that, may not thread safe.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/policy/FRAPolicy.java
rodrigoareis: That's a good point. I will think about that.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
rodrigoareis: good catch!

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAClient.java
rodrigoareis: Good point, the boolean as return is good to have, specially for the "hybrid" apps. I can remove the throw and catch this exception.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
rodrigoareis: I will remove nonCompliancePolicy attribute. Developers will have to rely on the logs or the Account.getLockingPolicy

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
rodrigoareis: No, copy & paste issue

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
rodrigoareis: Yes, the idea is to return the last registered. Return both mechanisms as result would require refactor the listener to FRAListener<Mechanism[]>, which would require a huge refactor across the SDK and Authenticator sub products

rodrigoareis: Well, this will be 4.0, so we maybe be able to return both we introduced a break change on FRAClient.createMechanismFromUri. I think it will be better

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
rodrigoareis: Not sure if I got it.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/Account.java
rodrigoareis: We want this treated as JSON String to make it easy to serialize to other platforms.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/Account.java
rodrigoareis: We want this treated as JSON String to make it easy to serialize to other platforms.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/Account.java
rodrigoareis: No, this should be "protected", we need to fix it

forgerock-authenticator/src/main/java/org/forgerock/android/auth/Account.java
```diff
@@ -153,6 +168,47 @@ public Calendar getTimeAdded() {
         return timeAdded;
     }
 
+    /**
+     * Return the Policies used to enforce App security.
+     * @return Policies in a JSON String format.
+     */
+    public String getPolicies() {
```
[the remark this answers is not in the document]
rodrigoareis: Yes, that would, but we want to make it easy to serialize to other platforms.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/MechanismParser.java
rodrigoareis: Done

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAPolicyEvaluator.java
rodrigoareis: Done, created a Result object to replace boolean as result

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAPolicyEvaluator.java
rodrigoareis: Done

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAClient.java
rodrigoareis: Done

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
rodrigoareis: Done, created a Result object to replace boolean as result

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
rodrigoareis: Done

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
rodrigoareis: Done

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
witrisna: The getPolicies() return a string of json, we check if the json contains a policy name, it may be easy to get name collision, for example, if developer create a policy called **Score**, that may collide with the jailbreakScore.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAPolicyEvaluator.java
witrisna: Looks like it will replace the existing list with withPolicies(), should it add to existing list?
FRAPolicyEvaluatorBuilder.withPolicy(A).withPolicy(B).withPolicy([C,D]), A and B will be removed.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/Account.java
witrisna: Sounds good

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
witrisna: The idea is to try best to register? if OATH failed, do we still want to proceed with the Push?

forgerock-core/src/main/java/org/forgerock/android/auth/biometric/BiometricAuth.kt
```diff
@@ -216,6 +219,40 @@ class BiometricAuth @JvmOverloads constructor(
...
+            var canAuthenticate = true
+            if (Build.VERSION.SDK_INT < 28) {
+                val keyguardManager : KeyguardManager = applicationContext.getSystemService(KEYGUARD_SERVICE) as KeyguardManager
+                val packageManager : PackageManager = applicationContext.packageManager
+                // Check if Fingerprint Sensor is supported
+                if(!packageManager.hasSystemFeature(PackageManager.FEATURE_FINGERPRINT)) {
+                    canAuthenticate = false
+                }
+                // Check if lock screen security is enabled in Settings
+                if (!keyguardManager.isKeyguardSecure) {
```
jeyanthanperiyasamy, line 234: keyguardManager.isKeyguardSecure.not()

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
rodrigoareis: Got it, good catch! I will fix it

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
rodrigoareis: Yes, that's what I got from Product. Previously, if anything fail, the registration was not completed.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
rodrigoareis: Done

forgerock-authenticator/src/main/java/org/forgerock/android/auth/FRAPolicyEvaluator.java
rodrigoareis: Yes, that's what is the javadoc of withPolicies `Sets a list of policies to be evaluated. This method overwrite an existing list.`

So, the developer should use something like this instead: 

```
FRAPolicyEvaluator.builder()
    .withPolicies([C,D])
    .withPolicy(A)
    .withPolicy(B)
```

witrisna (dismissed):

forgerock-core/src/main/java/org/forgerock/android/auth/biometric/BiometricAuth.kt
```diff
@@ -216,6 +219,40 @@ class BiometricAuth @JvmOverloads constructor(
...
         private val TAG = BiometricAuth::class.java.simpleName
+
+        @JvmStatic
+        fun isBiometricAvailable(applicationContext: Context) : Boolean{
+            var canAuthenticate = true
+            if (Build.VERSION.SDK_INT < 28) {
+                val keyguardManager : KeyguardManager = applicationContext.getSystemService(KEYGUARD_SERVICE) as KeyguardManager
+                val packageManager : PackageManager = applicationContext.packageManager
+                // Check if Fingerprint Sensor is supported
+                if(!packageManager.hasSystemFeature(PackageManager.FEATURE_FINGERPRINT)) {
```
jeyanthanperiyasamy, line 230: same here if(packageManager.hasSystemFeature(PackageManager.FEATURE_FINGERPRINT).not())

jeyanthanperiyasamy (commented):
just make sure since we lifted the RootDetector to FRCode ...just do a sanity test and make sure all the tests passed in CI/CD

spetrov (approved):
Looks really good to me! 👍🏻

jeyanthanperiyasamy (approved):

witrisna (dismissed):

spetrov (approved):

PR #261, opened 2023-05-23
SDKS-2509 Cancelling WebAuthn Key Selection dialog result with NotAllowedError, simulate as canceling WebAuthn Authentication.

witrisna:
# JIRA Ticket

[SDKS-2509](https://bugster.forgerock.org/jira/browse/SDKS-2509)

# Description

Cancelling WebAuthn Key Selection dialog result with NotAllowedError, simulate as canceling WebAuthn Authentication.


jeyanthanperiyasamy (approved):

forgerock-auth/src/main/java/org/forgerock/android/auth/webauthn/WebAuthKeySelectionFragment.kt
```diff
@@ -38,17 +40,10 @@ class WebAuthKeySelectionFragment : DialogFragment() {
...
-            field?.invokeOnCancellation {
-                if (this.isVisible) {
-                    dismiss()
-                }
-            }
-        }
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
+        isCancelable = false
```
jeyanthanperiyasamy, line 46: just a question . so if you set this , user cannot tap outside the dialog, they have to explicitly press the Cancel button. is that right ?

PR #268, opened 2023-06-29
Add ability to manually refresh access token

MAKARD:
# JIRA Ticket

N/A

# Description

Added an ability manually refresh access token regardless of its lifetime. 
Conforms to the same feature on iOS SDK within PR https://github.com/ForgeRock/forgerock-ios-sdk/pull/219

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/FRUser.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/FRUser.java
@@ -96,6 +96,19 @@ public class FRUser {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jeyanthanperiyasamy:
Tests required for this change

MAKARD:
> Tests required for this change

I would really appreciate if you could do that instead of me, ‘cause I’m not android developer really

george-bafaloukas-forgerock:
Hi @MAKARD  I think it is best if we close the PR and integrate the changes in a branch originated internally, so that we can sort the tests and the pipelines.
Thanks a lot for your effort and raising this PR.
We will keep you updated on the next steps.

Kind regards,
George Bafaloukas

PR #289, opened 2023-07-24
SDKS-2617 Revoke old access token upon force refresh

witrisna:
# JIRA Ticket

https://bugster.forgerock.org/jira/browse/SDKS-2617

# Description

Revoke old access token upon force refresh

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/DefaultTokenManager.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/DefaultTokenManager.java
@@ -164,6 +164,10 @@ class DefaultTokenManager implements TokenManager {
             Listener.onException(listener, new AuthenticationRequiredException("Refresh Token does not exists."));
             return;
         }
+        //If the access token is not expired yet, revoke it
+        if (!accessToken.isExpired()) {
+            oAuth2Client.revoke(accessToken, false, null);
+        }
         Logger.debug(TAG, "Exchange AccessToken with Refresh Token");
         oAuth2Client.refresh(accessToken.getSessionToken(), refreshToken, new FRListener<AccessToken>() {
             @Override
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/OAuth2Client.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/OAuth2Client.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 - 2022 ForgeRock. All rights reserved.
+ * Copyright (c) 2019 - 2023 ForgeRock. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -212,12 +212,25 @@ public class OAuth2Client {
      * @param listener    Listener to listen for revoke event
      */
     public void revoke(@NonNull AccessToken accessToken, final FRListener<Void> listener) {
+        revoke(accessToken, true, listener);
+    }
+
+    /**
+     * Revoke the AccessToken, to revoke the access token, first look for refresh token to revoke, if
+     * not provided or useRefreshToken = false, will revoke with the access token.
+     *
+     * @param accessToken The AccessToken to be revoked
+     * @param useRefreshToken If true, revoke with refresh token, otherwise revoke access token
+     * @param listener    Listener to listen for revoke event
+     */
+    public void revoke(@NonNull AccessToken accessToken, boolean useRefreshToken, final FRListener<Void> listener) {
         Logger.debug(TAG, "Revoking Access Token & Refresh Token");
         final OAuth2ResponseHandler handler = new OAuth2ResponseHandler();
         try {
             FormBody.Builder builder = new FormBody.Builder();
 
-            String token = accessToken.getRefreshToken() == null ? accessToken.getValue() : accessToken.getRefreshToken();
+            String token = accessToken.getRefreshToken() == null || !useRefreshToken
+                    ? accessToken.getValue() : accessToken.getRefreshToken();
 
             RequestBody body = builder
                     .add(OAuth2.CLIENT_ID, clientId)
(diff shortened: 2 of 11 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/DefaultTokenManager.java
```diff
@@ -164,6 +164,10 @@ public void refresh(@NonNull AccessToken accessToken, final FRListener<AccessTok
             Listener.onException(listener, new AuthenticationRequiredException("Refresh Token does not exists."));
             return;
         }
+        //If the access token is not expired yet, revoke it
+        if (!accessToken.isExpired()) {
```
jeyanthanperiyasamy, line 168: is this check required ?  what happens when you try to revoke the expired token .  i had a android device in the past which my system time is always wrong due to calendar issues. so will there be a situation like server token is not expired and the client is expired with wrong calendar . its a edge case .

jeyanthanperiyasamy (commented):
looks good , just a minor comment

jeyanthanperiyasamy (dismissed):

forgerock-auth/src/main/java/org/forgerock/android/auth/DefaultTokenManager.java
witrisna: Server returns the token "Expire In" which is the duration for the token, but not the exact time when the token will expired, client-side keep track of when the token expires, however, if the user change device time after receiving the token, the edge case scenario may happen. 

I put the logic to revoke the access token within the defined threshold, and avoid making additional network calls with the expired token. For e.g, if the access token has 30 min lifetime, and user set 10mins threshold, the access token still valid for 10mins.

forgerock-auth/src/main/java/org/forgerock/android/auth/OAuth2Client.java
```diff
@@ -211,13 +211,14 @@ public void onResponse(@NotNull Call call, @NotNull Response response) {
      * @param accessToken The AccessToken to be revoked
      * @param listener    Listener to listen for revoke event
      */
-    public void revoke(@NonNull AccessToken accessToken, final FRListener<Void> listener) {
```
rodrigoareis, line 214: Update copyright

forgerock-auth/src/main/java/org/forgerock/android/auth/OAuth2Client.java
```diff
@@ -211,13 +211,14 @@ public void onResponse(@NotNull Call call, @NotNull Response response) {
...
+    public void revoke(@NonNull AccessToken accessToken, boolean useRefreshToken, final FRListener<Void> listener) {
```
rodrigoareis, line 214: Update javadoc to include new `useRefreshToken` parameter. Does the new parameter introduces a break change?

forgerock-auth/src/main/java/org/forgerock/android/auth/SessionManager.java
```diff
@@ -44,8 +45,17 @@ public SessionManager(TokenManager tokenManager, SingleSignOnManager singleSignO
     }
 
     @WorkerThread
-    public void refresh(AccessToken accessToken, FRListener<AccessToken> listener) throws AuthenticationRequiredException {
-        tokenManager.refresh(accessToken, listener);
+    public void refresh(FRListener<AccessToken> listener) {
```
rodrigoareis, line 48: Should we include javadoc here? I see more missing javadocs for other methods in this class

forgerock-auth/src/test/java/org/forgerock/android/auth/DefaultTokenManagerTest.kt
```diff
@@ -266,6 +266,7 @@ class DefaultTokenManagerTest : BaseTest() {
     @Test
     @Throws(Throwable::class)
     fun testTokenRefreshWithThreshold() {
+        server.enqueue(MockResponse().setResponseCode(HttpURLConnection.HTTP_OK))
```
rodrigoareis, line 269: Update copyright

rodrigoareis (changes requested):
Overall changes looks good to me. Minor comments to be reviewed

forgerock-auth/src/main/java/org/forgerock/android/auth/SessionManager.java
witrisna: Javadoc added.

forgerock-auth/src/test/java/org/forgerock/android/auth/DefaultTokenManagerTest.kt
witrisna: Updated

forgerock-auth/src/main/java/org/forgerock/android/auth/TokenManager.java
```diff
@@ -50,9 +50,8 @@ public interface TokenManager {
      *
      * @param accessToken AccessToken
      * @param listener    Listener to listen for refresh event.
-     * @throws AuthenticationRequiredException When failed to Refresh the {@link AccessToken}
      */
-    void refresh(AccessToken accessToken, FRListener<AccessToken> listener) throws AuthenticationRequiredException;
+    void refresh(AccessToken accessToken, FRListener<AccessToken> listener);
```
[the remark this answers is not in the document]
witrisna: Updated

forgerock-auth/src/main/java/org/forgerock/android/auth/OAuth2Client.java
witrisna: Javadoc added

forgerock-auth/src/main/java/org/forgerock/android/auth/OAuth2Client.java
witrisna: Updated

rodrigoareis (approved):
Changes looks good to me

spetrov (approved):
Looks good! 👍🏻

PR #300, opened 2023-07-28
SDKS 2576 - WebAuthn Lost credentials

jeyanthanperiyasamy:
(https://bugster.forgerock.org/jira/browse/SDKS-2576)

https://forgerock.zendesk.com/agent/tickets/94902

### DIFF
--- a/forgerock-core/src/main/java/org/forgerock/android/auth/EncryptedSharedPreferences.kt
+++ b/forgerock-core/src/main/java/org/forgerock/android/auth/EncryptedSharedPreferences.kt
@@ -15,7 +15,6 @@ import androidx.security.crypto.MasterKey
 import java.io.File
 import java.security.KeyStore
 
-
 class EncryptedPreferences {
     companion object {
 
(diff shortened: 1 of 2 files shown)


forgerock-core/src/main/java/org/forgerock/android/auth/EncryptedSharedPreferences.kt
```diff
@@ -67,22 +83,45 @@ class EncryptedPreferences {
...
                 fileName,
-                masterKeyAlias,
+                builder.build(),
                 EncryptedSharedPreferences.PrefKeyEncryptionScheme.AES256_SIV,
                 EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM
             )
         }
+
+        // master key recovery code
+        private fun recoverFromDefaultMasterKey(
```
witrisna, line 104: May provide javadoc, that this method is to extract data which encrypted with MasterKey, and delete the master key.

forgerock-core/src/main/java/org/forgerock/android/auth/EncryptedSharedPreferences.kt
```diff
@@ -67,22 +83,45 @@ class EncryptedPreferences {
...
+            context: Context,
+            fileName: String = "secret_shared_prefs" + context.packageName
+        ): MutableMap<String, MutableSet<String>> {
+            val cache = mutableMapOf<String, MutableSet<String>>()
+            // creating all 1.0.5 alpha files
+            try {
+                val secureSharedPreferences =
+                    createPreferencesFile(context, fileName)
+                secureSharedPreferences.all.entries.map { it.key }.forEach { key ->
+                    secureSharedPreferences.getStringSet(key, emptySet())?.let {
```
witrisna, line 114: Just to confirm, we previously set data as StringSet?

jeyanthanperiyasamy: yes , this is how we set it before in the WebAuthnDataRepository

```       encryptedPreferences.edit().apply {
                map.forEach {
                    this.putStringSet(it.key, it.value)
                }
            }.apply()
```

spetrov (approved):
👍🏻

PR #354, opened 2023-11-03
SDKS-2797 Switch to use auth-per-use Key for Device Binding

witrisna:
# JIRA Ticket

[SDKS-2797](https://bugster.forgerock.org/jira/browse/SDKS-2797)

# Description

Switch to use auth-per-use Key for Device Binding

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
@@ -245,6 +245,7 @@ open class DeviceBindingCallback : AbstractCallback, Binding {
                             deviceBindingRepository.persist(it)
                             val jws = deviceAuthenticator.sign(context,
                                 kp,
+                                status.signature,
                                 it.kid,
                                 userId,
                                 challenge,
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricAndDeviceCredential.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricAndDeviceCredential.kt
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2023 ForgeRock. All rights reserved.
+ *
+ *  This software may be modified and distributed under the terms
+ *  of the MIT license. See the LICENSE file for details.
+ */
+
+package org.forgerock.android.auth.devicebind
+
+import android.annotation.SuppressLint
+import android.content.Context
+import android.os.Build
+import android.security.keystore.KeyProperties
+import androidx.biometric.BiometricManager
+import androidx.biometric.BiometricPrompt
+import org.forgerock.android.auth.callback.Attestation
+import org.forgerock.android.auth.callback.DeviceBindingAuthenticationType
+import java.security.PrivateKey
+import java.security.interfaces.RSAPublicKey
+
+/**
+ * Settings for all the biometric authentication and device credential is configured
+ */
+open class BiometricAndDeviceCredential : BiometricAuthenticator() {
+    override fun authenticate(authenticationCallback: BiometricPrompt.AuthenticationCallback,
+                              privateKey: PrivateKey) {
+        biometricInterface.authenticate(authenticationCallback)
+    }
+
+    /**
+     * generate the public and private keypair
+     */
+    @SuppressLint("NewApi")
+    override suspend fun generateKeys(context: Context, attestation: Attestation): KeyPair {
+        val builder = cryptoKey.keyBuilder()
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            builder.setAttestationChallenge(attestation.challenge)
+        }
+
+        //We use time-base key because we allow device credential as fallback
+        //Device credential is not consider biometric strong
+        if (isApi30OrAbove) {
+            builder.setUserAuthenticationParameters(cryptoKey.timeout,
+                KeyProperties.AUTH_BIOMETRIC_STRONG or KeyProperties.AUTH_DEVICE_CREDENTIAL)
+        } else {
+            builder.setUserAuthenticationValidityDurationSeconds(cryptoKey.timeout)
+        }
+
+        builder.setUserAuthenticationRequired(true)
+        val key = cryptoKey.createKeyPair(builder.build())
+        return KeyPair(key.public as RSAPublicKey, key.private, cryptoKey.keyAlias)
+    }
+
+    /**
+     * check biometric is supported
+     */
+    override fun isSupported(context: Context, attestation: Attestation): Boolean {
+        return super.isSupported(context, attestation) &&
+                biometricInterface.isSupported(
+                    BiometricManager.Authenticators.BIOMETRIC_STRONG or BiometricManager.Authenticators.DEVICE_CREDENTIAL,
+                    BiometricManager.Authenticators.BIOMETRIC_WEAK or BiometricManager.Authenticators.DEVICE_CREDENTIAL)
+    }
+
+
+    final override fun type(): DeviceBindingAuthenticationType =
+        DeviceBindingAuthenticationType.BIOMETRIC_ALLOW_FALLBACK
+
+
+}
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricAuthenticator.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricAuthenticator.kt
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2023 ForgeRock. All rights reserved.
+ *
+ *  This software may be modified and distributed under the terms
+ *  of the MIT license. See the LICENSE file for details.
+ */
+
+package org.forgerock.android.auth.devicebind
+
+import android.content.Context
+import android.os.Build
+import androidx.annotation.VisibleForTesting
+import androidx.biometric.BiometricPrompt
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import org.forgerock.android.auth.CryptoKey
+import java.security.PrivateKey
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
+
+abstract class BiometricAuthenticator : CryptoAware, DeviceAuthenticator {
+
+    @VisibleForTesting
+    internal var isApi30OrAbove = Build.VERSION.SDK_INT >= Build.VERSION_CODES.R
+    internal lateinit var cryptoKey: CryptoKey
+    internal lateinit var biometricInterface: BiometricHandler
+
+    final override fun setKey(cryptoKey: CryptoKey) {
+        this.cryptoKey = cryptoKey
+    }
+
+    fun setBiometricHandler(biometricHandler: BiometricHandler) {
+        this.biometricInterface = biometricHandler
+    }
+
+    override fun deleteKeys(context: Context) {
+        cryptoKey.deleteKeys()
+    }
+
+    /**
+     * Display biometric prompt for authentication type
+     * @param context Application Context
+     * @return statusResult Listener for receiving Biometric changes
+     */
+    override suspend fun authenticate(context: Context): DeviceBindingStatus =
+        withContext(Dispatchers.Main) {
+            suspendCoroutine { continuation ->
+                //The keys may be removed due to pin change
+                val privateKey = cryptoKey.getPrivateKey()
+                if (privateKey == null) {
+                    continuation.resume(DeviceBindingErrorStatus.ClientNotRegistered())
+                } else {
+                    val listener = object : BiometricPrompt.AuthenticationCallback() {
+
+                        override fun onAuthenticationError(errorCode: Int,
+                                                           errString: CharSequence) {
+                            when (errorCode) {
+                                BiometricPrompt.ERROR_CANCELED, BiometricPrompt.ERROR_USER_CANCELED, BiometricPrompt.ERROR_NEGATIVE_BUTTON -> continuation.resume(
+                                    DeviceBindingErrorStatus.Abort(errString.toString(),
+                                        code = errorCode))
+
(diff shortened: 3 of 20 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricAndDeviceCredential.kt
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        //We use time-base key because we allow device credential as fallback
+        //Device credential is not consider biometric strong
+        if (isApi30OrAbove) {
+            builder.setUserAuthenticationParameters(cryptoKey.timeout,
+                KeyProperties.AUTH_BIOMETRIC_STRONG or KeyProperties.AUTH_DEVICE_CREDENTIAL)
+        } else {
+            builder.setUserAuthenticationValidityDurationSeconds(cryptoKey.timeout)
+        }
```
witrisna, line 47: For BiometricAndDeviceCredential, we keep using the time-based key, because the Device Credential is not consider strong Biometric.

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricOnly.kt
```diff
@@ -0,0 +1,86 @@
...
+                              privateKey: PrivateKey) {
+        try {
+            biometricInterface.authenticate(authenticationCallback,
+                BiometricPrompt.CryptoObject(getSignature(privateKey)))
+        } catch (e: Exception) {
+            //Failed because the key was generated with
+            //KeyGenParameterSpec.Builder.setUserAuthenticationParameters
+            Logger.warn(TAG, "Fallback to time-based key", e)
+            biometricInterface.authenticate(authenticationCallback)
+        }
```
witrisna, line 46: The Key may be created with time-based key, handling fallback.

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricOnly.kt
```diff
@@ -0,0 +1,86 @@
...
+     * generate the public and private keypair
+     */
+    @SuppressLint("NewApi")
+    override suspend fun generateKeys(context: Context, attestation: Attestation): KeyPair {
+        val builder = cryptoKey.keyBuilder()
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            builder.setAttestationChallenge(attestation.challenge)
+        }
+
+        if (biometricInterface.isSupportedBiometricStrong().not()) {
```
witrisna, line 59: Only use auth-per-use if the device support strong biometric

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindAuthenticators.kt
```diff
@@ -79,6 +56,15 @@ interface DeviceAuthenticator {
         //Do Nothing
     }
 
+    /**
+     * The JWS algorithm (alg) parameter.
+     * Header Parameter identifies the cryptographic algorithm
+     * used to secure the JWS.
+     */
+    fun getAlgorithm(): String {
```
witrisna, line 64: Allow developer to override the sign algorithm

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindAuthenticators.kt
```diff
@@ -108,7 +98,15 @@ interface DeviceAuthenticator {
                 .claim(PLATFORM, "android")
                 .claim(ANDROID_VERSION, Build.VERSION.SDK_INT)
                 .claim(CHALLENGE, challenge).build())
-        signedJWT.sign(RSASSASigner(keyPair.privateKey))
+        signature?.let {
```
witrisna, line 101: If signature is provided, we use the signature to sign, otherwise use private key.

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindingStatus.kt
```diff
@@ -55,7 +56,13 @@ abstract class DeviceBindingErrorStatus(var message: String,
...
 }
 
-data class Success(val privateKey: PrivateKey) : DeviceBindingStatus
+/**
+ * Represent the success status after [DeviceAuthenticator.authenticate]
+ *
+ * @property privateKey The unlocked private key
+ * @property signature The unlocked signature
+ */
+data class Success(val privateKey: PrivateKey, val signature: Signature? = null) : DeviceBindingStatus
```
witrisna, line 65: Signature is optional, only available when using auth-per-use key

forgerock-core/src/main/java/org/forgerock/android/auth/CryptoKey.kt
```diff
@@ -30,7 +30,7 @@ class CryptoKey(private var keyId: String) {
     //For hashing the keyId
     private val hashingAlgorithm = "SHA-256"
     val keySize = 2048
-    val timeout = 60
+    val timeout = 5
```
witrisna, line 33: Even with time-based key, we sign it immediately, shorter the window to reduce the risk.

forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
```diff
@@ -245,6 +245,7 @@ open class DeviceBindingCallback : AbstractCallback, Binding {
                             deviceBindingRepository.persist(it)
                             val jws = deviceAuthenticator.sign(context,
                                 kp,
+                                status.signature,
```
witrisna, line 248: An unlocked signature is added to the interface. We need to use the same instance from CryptoObject to the Signing.

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/RSASASignatureSigner.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.nimbusds.jose.util.Base64URL
+import java.security.InvalidKeyException
+import java.security.PrivateKey
+import java.security.Signature
+import java.security.SignatureException
+
+/**
+ * Extend of [RSASSASigner] to cache and expose the signature
+ */
+internal class RSASASignatureSigner : RSASSASigner {
```
witrisna, line 24: Has to extend the original implementation of RSASASigner, RSASASigner does not allow to assess the signature which required for the CryptoObject.

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricAuthenticator.kt
```diff
@@ -0,0 +1,116 @@
...
+
+                                else -> {
+                                    continuation.resume(DeviceBindingErrorStatus.Unknown(errString.toString(),
+                                        code = errorCode))
+                                }
+                            }
+                        }
+
+                        override fun onAuthenticationSucceeded(result: BiometricPrompt.AuthenticationResult) {
+                            result.cryptoObject?.let {
```
jeyanthanperiyasamy, line 89: i think we can do one line 

  result.cryptoObject?.signature?.let {
                                continuation.resume(Success(privateKey, it))
                                return
                            }

jeyanthanperiyasamy:
i am not finished this yet, will continue later

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricAndDeviceCredential.kt
jeyanthanperiyasamy: yes this make sense ,. should we update the docs mentioning what keys are used by different authenticator type? 

for example: BiometricOnly uses authBased , BiometricAndDeviceCredential uses timebased

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricBindingHandler.kt
```diff
@@ -86,17 +89,14 @@ internal class BiometricBindingHandler(titleValue: String,
             when {
                 // API 29 and above, check the support for STRONG type
                 this.hasBiometricCapability(strongAuthenticators) -> {
-                    this.authenticatorType = AuthenticatorType.STRONG
```
jeyanthanperiyasamy, line 89: so these are the side effects?

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricBindingHandler.kt
```diff
@@ -107,4 +107,15 @@ internal class BiometricBindingHandler(titleValue: String,
         }
         return false
     }
+
+    override fun isSupportedBiometricStrong(): Boolean {
+        biometricAuth?.apply {
+            when {
+                this.hasBiometricCapability(BIOMETRIC_STRONG) -> {
```
jeyanthanperiyasamy, line 114: we can simplify this to one line code
 
       return biometricAuth?.hasBiometricCapability(BIOMETRIC_STRONG) ?: false

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindAuthenticators.kt
jeyanthanperiyasamy: may be we need this change in iOS also ? we need to double check that . i am sure we use EC, will need to look can we support RSA also

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/RSASASignatureSigner.kt
jeyanthanperiyasamy, line 24: can we write some unit test for this class ?

forgerock-core/src/main/java/org/forgerock/android/auth/biometric/BiometricAuth.kt
```diff
@@ -67,12 +70,8 @@ class BiometricAuth @JvmOverloads constructor(
      */
     private val description: String? = null,
 
-    /**
-     * The type of authenticator to be displayed.
-     * @return the authenticatorType as Int.
-     */
-    var authenticatorType: AuthenticatorType = AuthenticatorType.WEAK
```
jeyanthanperiyasamy, line 74: we need to test the authenticator app and make sure we didnt break the biometric after user receive the push notification . probably i will ask stoyan to do this test. OR i can do this as well

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricAndDeviceCredential.kt
witrisna: For now, I don't think it is necessary to get to that detail.

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindAuthenticators.kt
witrisna: We did some research before, and iOS only supports EC if I remember correctly. If that is the case, it may not be necessary for iOS to override. I think the client may want to change the alg for performance improvement.

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricBindingHandler.kt
witrisna: Yes, this is the side effect I mentioned, the method called "isSupported", it sounds like only checks if the device supports it, however, it changes state as well.

forgerock-core/src/main/java/org/forgerock/android/auth/biometric/BiometricAuth.kt
witrisna: I agree, this code is share with Authenticator App.

jeyanthanperiyasamy (commented):
provided comments

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/RSASASignatureSigner.kt
```diff
@@ -0,0 +1,61 @@
...
+    var signature: Signature
+        private set
+
+    constructor(jwsAlgorithm: JWSAlgorithm , privateKey: PrivateKey) : super(privateKey) {
+        signature = getInitiatedSignature(jwsAlgorithm)
```
jeyanthanperiyasamy, line 29: if you provide the private key the signature value is same for both binding and signing scenarios?

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/RSASASignatureSigner.kt
witrisna: Instead of extending the RSASSASigner, extend JWSSigner

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricAuthenticator.kt
witrisna: Updated

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/BiometricBindingHandler.kt
witrisna: Updated

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/RSASASignatureSigner.kt
witrisna: Unit test added

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/RSASASignatureSigner.kt
witrisna: Implementation updated, only take Signature.

spetrov (approved):
Looks good to me!

PR #389, opened 2024-01-26
SDKS-2935 Added test coverage the failure outcome in the DeviceSigningVerifier node

spetrov:
# JIRA Ticket

[SDKS-2935](https://bugster.forgerock.org/jira/browse/SDKS-2935): If a user is set to `inactive` the `DeviceSigningVerifier` node throws an exception

# Description
- Added test coverage for the `Capture Failure` option in the Device Signing Verifier node
- Added test coverage for the `DeviceSigningVerifierNode.FAILURE` shared state variable


witrisna (dismissed):
LGTM

PR #401, opened 2024-03-12
SDKS- 3033, Send header attribute identifying the SDK

jeyanthanperiyasamy:
# JIRA Ticket

Please link jira ticket here

# Description


forgerock-core/src/main/java/org/forgerock/android/auth/NetworkConfig.java
```diff
@@ -1,31 +1,30 @@
 /*
```
jeyanthanperiyasamy, line 1: i want to convert this to kotlin at some point in a different story .. but since i modified the OKHttpClientProvider to kotlin , i have to delombak this class

forgerock-core/src/main/java/org/forgerock/android/auth/OkHttpClientProvider.kt
```diff
@@ -0,0 +1,101 @@
...
+            networkConfig.interceptorSupplier.get() != null
+        ) {
+            for (i in networkConfig.interceptorSupplier.get()) {
+                builder.addInterceptor(i)
+            }
+        }
+
+        builder.addInterceptor(Interceptor { chain ->
+            val request = chain.request().newBuilder()
+                .header("x-requested-with", "forgerock-sdk")
```
rodrigoareis, line 59: Can we move those new strings to a constant?

rodrigoareis (dismissed):
Changes looks good to me

witrisna (approved):

rodrigoareis (approved):
Changes looks good to me

vahancouver (approved):
LGTM

spetrov (approved):

PR #402, opened 2024-03-14
SDKS-3015 SDK 4.4.0 Release Testing

spetrov:
# JIRA Ticket

[SDKS-3015](https://bugster.forgerock.org/jira/browse/SDKS-3015) SDK 4.4.0 Release Testing

# Description
Some of the device binding/signing e2e test cases were failing when running in parallel on multiple devices, because they were using the same user and some assumptions were not met - for example, some a test may make the user inactive, and at the same time, other tests assume/require that the same user is active...
For that reason, I introduced the following improvements: 
- All device binding/signing related test cases now create a brand new random user...
- Introduced a `RetryTestRule` annotation and added it to the device binding tests. This way a test would be repeated 3 times if it fails... This is useful for flaky tests...
- "Suppressed" the `KeyAttestation` e2e test cases on Android 6, since this feature is not available on this version... 
- Added `SkipTestOnPermissionFailureRule` to the device location-related tests. This helps in some cases and I see that this works as expected... Nevertheless, there are still rare cases when location data is missing (even though permissions seem to have been granted successfully)...

In general, these measures are a big improvement IMO, but more work is needed...
This is an example [test run](https://cloud.bitbar.com/#testing/device-session/207366434/211529522) on BitBar, with the proposed changes on all supported Android versions (6 to 14)... As you can see on Android 7 and 8 we still have some tests failing... :-( 

image

### DIFF
--- a/forgerock-integration-tests/src/androidTest/java/org/forgerock/android/auth/AndroidBaseTest.java
+++ b/forgerock-integration-tests/src/androidTest/java/org/forgerock/android/auth/AndroidBaseTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 ForgeRock. All rights reserved.
+ * Copyright (c) 2020 - 2024 ForgeRock. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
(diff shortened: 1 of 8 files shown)


forgerock-integration-tests/src/androidTest/java/org/forgerock/android/auth/AndroidBaseTest.java
```diff
@@ -33,4 +44,77 @@ public void setUpSDK() {
...
+                    node.next(context, this );
+                }
+                if (node.getCallback(KbaCreateCallback.class) != null) {
+                    List<Callback> callbacks = node.getCallbacks();
+
+                    KbaCreateCallback firstQuestion = (KbaCreateCallback) callbacks.get(0);
+                    firstQuestion.setSelectedQuestion(firstQuestion.getPredefinedQuestions().get(0));
+                    firstQuestion.setSelectedAnswer("Test");
+
+                    /*
```
jeyanthanperiyasamy, line 102: remove commented code ?

jeyanthanperiyasamy (approved):
mostly trying to reduce the flakyness in the test. Good work

forgerock-integration-tests/src/androidTest/java/org/forgerock/android/auth/callback/BaseDeviceBindingTest.java
```diff
@@ -47,6 +52,9 @@ public abstract class BaseDeviceBindingTest {
     @Rule
     public ExpectedException thrown = ExpectedException.none();
 
+    @Rule
+    public RetryTestRule retry = new RetryTestRule(3);
```
jeyanthanperiyasamy, line 56: we are introducing retry now ? test retries can be a blessing or a curse. we need to keep an eye on how much we do in all the test

jeyanthanperiyasamy (approved):
mostly trying to reduce the flakyness in the test. Good work

forgerock-integration-tests/src/androidTest/java/org/forgerock/android/auth/AndroidBaseTest.java
spetrov: Added a comment in the block

forgerock-integration-tests/src/androidTest/java/org/forgerock/android/auth/callback/BaseDeviceBindingTest.java
spetrov: Yes, I agree... I removed this retry from the device-binding tests...

PR #406, opened 2024-03-27
SDKS-3079 Replace deprecated onActivityResult with ActivityResultContract

witrisna:
# JIRA Ticket

[SDKS-3079](https://bugster.forgerock.org/jira/browse/SDKS-3079)

# Description

Replace deprecated onActivityResult with ActivityResultContract

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/AppAuthFragment2.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/AppAuthFragment2.kt
@@ -1,0 +1,96 @@
+/*
+ * Copyright (c) 2024 ForgeRock. All rights reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ */
+package org.forgerock.android.auth
+
+import android.content.Intent
+import android.os.Bundle
+import androidx.annotation.RestrictTo
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentActivity
+import androidx.fragment.app.FragmentManager
+import kotlinx.coroutines.flow.MutableStateFlow
+import net.openid.appauth.AuthorizationResponse
+import org.forgerock.android.auth.centralize.BrowserLauncher
+import org.forgerock.android.auth.centralize.Launcher
+
+private const val PENDING = "pending"
+
+/**
+ * Headless Fragment to receive callback result from AppAuth library
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY)
+class AppAuthFragment2 : Fragment() {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        val state: MutableStateFlow<Intent?> = MutableStateFlow(null)
+        val delegate =
+            registerForActivityResult(AuthorizeContract()) {
+                state.value = it
+            }
+
+        val pending = savedInstanceState?.getBoolean(PENDING, false) ?: false
+
+        BrowserLauncher.init(Launcher(delegate, state, pending))
+    }
+
+    override fun onSaveInstanceState(outState: Bundle) {
+        outState.putBoolean(PENDING, true)
+        super.onSaveInstanceState(outState)
+    }
+
+    override fun onDestroy() {
+        super.onDestroy()
+        BrowserLauncher.reset()
+    }
+
+    companion object {
+        val TAG: String = AppAuthFragment2::class.java.name
+
+        /**
+         * Initialize the Fragment to receive AppAuth callback event.
+         */
+        @Synchronized
+        @JvmStatic
+        fun launch(activity: FragmentActivity, browser: FRUser.Browser) {
+            val fragmentManager: FragmentManager = activity.supportFragmentManager
+            var current = fragmentManager.findFragmentByTag(TAG) as? AppAuthFragment2
+            if (current == null) {
+                current = AppAuthFragment2()
+                fragmentManager.beginTransaction().add(current, TAG).commitNow()
+            }
+
+            BrowserLauncher.authorize(browser, object : FRListener<AuthorizationResponse> {
+                override fun onSuccess(result: AuthorizationResponse) {
+                    reset(activity, current)
+                    browser.listener.onSuccess(result)
+                }
+
+                override fun onException(e: Exception) {
+                    reset(activity, current)
+                    browser.listener.onException(e)
+                }
+
+                /**
+                 * Once receive the result, reset state.
+                 */
+                private fun reset(activity: FragmentActivity, fragment: Fragment?) {
+                    activity.runOnUiThread {
+                        BrowserLauncher.reset()
+                    }
+                    fragment?.let {
+                        activity.runOnUiThread {
+                            fragmentManager.beginTransaction().remove(it).commitNow()
+                        }
+                    }
+                }
+            })
+
+
+        }
+    }
+}
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/FRUser.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/FRUser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 - 2023 ForgeRock. All rights reserved.
+ * Copyright (c) 2019 - 2024 ForgeRock. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -14,15 +14,16 @@ import android.content.pm.PackageManager;
 import android.content.pm.ResolveInfo;
 import android.net.Uri;
 
+import androidx.annotation.NonNull;
 import androidx.annotation.VisibleForTesting;
 import androidx.annotation.WorkerThread;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentActivity;
-import androidx.fragment.app.FragmentManager;
 
 import net.openid.appauth.AuthorizationResponse;
 import net.openid.appauth.RedirectUriReceiverActivity;
 
+import org.forgerock.android.auth.centralize.BrowserLauncher;
 import org.forgerock.android.auth.exception.AlreadyAuthenticatedException;
 import org.forgerock.android.auth.exception.AuthenticationRequiredException;
 import org.forgerock.android.auth.exception.InvalidRedirectUriException;
@@ -31,8 +32,6 @@ import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
 
-import lombok.AccessLevel;
-import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 
 public class FRUser {
@@ -99,7 +98,7 @@ public class FRUser {
      * Refresh the {@link AccessToken} asynchronously, force token refresh, no matter the stored {@link AccessToken} is expired or not
      * refresh the token and persist it.
      *
-     * @param listener    Listener to listen for refresh event.
+     * @param listener Listener to listen for refresh event.
      */
     @WorkerThread
     public void refresh(FRListener<AccessToken> listener) {
@@ -256,10 +255,10 @@ public class FRUser {
         return new Browser();
     }
 
-    @Getter(AccessLevel.PACKAGE)
     public static class Browser {
 
-        public FRListener<AuthorizationResponse> listener;
+        private static final String TAG = Browser.class.getName();
+        private FRListener<AuthorizationResponse> listener;
         private AppAuthConfigurer appAuthConfigurer = new AppAuthConfigurer(this);
         private boolean failedOnNoBrowserFound = true;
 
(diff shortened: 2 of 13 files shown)


samples/app/src/main/java/com/example/app/centralize/CentralizeLoginViewModel.kt
spetrov, line 1: Update copyright

spetrov (changes requested):
Looks good to me. Have 2 minor comments. 
Please, update the changelog too...

forgerock-auth/src/main/java/org/forgerock/android/auth/OAuth2Client.java
```diff
@@ -459,4 +455,23 @@ private PKCE generateCodeChallenge() throws UnsupportedEncodingException {
         }
     }
 
+    public String getClientId() {
```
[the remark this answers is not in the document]
witrisna: Update the interface to package only, so it is not expose to developer.

samples/app/src/main/java/com/example/app/centralize/CentralizeLoginViewModel.kt
witrisna: Updated

spetrov (dismissed):

forgerock-auth/src/main/java/org/forgerock/android/auth/FRUser.java
```diff
@@ -317,35 +305,38 @@ private void login(Context context, FragmentManager manager, FRListener<FRUser>
             }
 
             try {
-                validateRedirectUri(context);
+                validateRedirectUri(activity.getApplicationContext());
```
jeyanthanperiyasamy, line 308: i think we can just give validateRedirectUri(activity)

forgerock-auth/src/main/java/org/forgerock/android/auth/FRUser.java
```diff
@@ -298,17 +297,6 @@ public void login(Fragment fragment, FRListener<FRUser> listener) {
          *                 <b> throws {@link java.net.MalformedURLException} When failed to parse the URL for API request.
          */
         public void login(FragmentActivity activity, FRListener<FRUser> listener) {
```
jeyanthanperiyasamy, line 299: For compose project , i think people mix and match ComponentActivity or AppCompatActivity

(AppCompatActivity extends FragmentActivity extends ComponentActivity)

So , Do you think we can expose a new method -  public void login(ComponentActivity activity, FRListener<FRUser> listener)

forgerock-auth/src/main/java/org/forgerock/android/auth/centralize/BrowserLauncher.kt
```diff
@@ -0,0 +1,88 @@
...
+    @Synchronized
+    internal fun init(launcher: Launcher) {
+        BrowserLauncher.launcher = launcher
+    }
+
+    /**
+     * reset the Launcher state
+     */
+    @Synchronized
+    internal fun reset() {
```
jeyanthanperiyasamy, line 43: who is using this ? cant find the caller

forgerock-auth/src/main/java/org/forgerock/android/auth/FRUser.java
```diff
@@ -281,7 +280,7 @@ public AppAuthConfigurer appAuthConfigurer() {
          *                 <b> throws {@link java.net.MalformedURLException} When failed to parse the URL for API request.
          */
         public void login(Fragment fragment, FRListener<FRUser> listener) {
```
jeyanthanperiyasamy, line 282: an another idea would be like , we can expose third parameter in login method like a Bool or something. if the user sets that value we can init the registerBrowserLauncher() or fallback to old route

jeyanthanperiyasamy:
overall i like this much better approach.. we should not depend on the Fragment to launch the centralize login. if the fragment got killed  or recreated by android system then it will be in the zombie state

forgerock-auth/src/main/java/org/forgerock/android/auth/FRUser.java
witrisna: The function registerForActivityResult() requires to run inside onCreate, after creation, when calling registerForActivityResult will throw an exception, the method registerBrowserLauncher -> registerForActivityResult, so it has to run inside onCreate(). I tried not to change the existing interface, and typically App Activity should either extend AppCompatActivity or FragmentActivity.

forgerock-auth/src/main/java/org/forgerock/android/auth/centralize/BrowserLauncher.kt
witrisna: this call androidx.activity.result.ActivityResultLauncher#unregister
```
Unregisters this launcher, releasing the underlying result callback, and any references captured within it. You should call this if the registry may live longer than the callback registered for this launcher.
```
I did not implement this in our Sample App, because our App has only one Activity, but I can add it back to the Activity.

forgerock-auth/src/main/java/org/forgerock/android/auth/FRUser.java
witrisna: We cannot call registerBrowserLauncher in login due to Activity Lifecycle.

forgerock-auth/src/main/java/org/forgerock/android/auth/FRUser.java
witrisna: Updated.

forgerock-auth/src/main/java/org/forgerock/android/auth/centralize/BrowserLauncher.kt
witrisna: Sample app updated.

forgerock-auth/src/main/java/org/forgerock/android/auth/centralize/BrowserLauncher.kt
```diff
@@ -0,0 +1,96 @@
...
+            return true
+        } ?: false
+    }
+}
+
+/**
+ * Register the Browser Launcher
+ */
+fun ComponentActivity.registerBrowserLauncher() {
+    val state: MutableStateFlow<Intent?> = MutableStateFlow(null)
```
jeyanthanperiyasamy, line 79: i think if we use MutableSharedFlow<Intent>() , then you can avoid this default value i belive and dont have to drop the first value

witrisna: Some information about best practice of using Flow Type

https://medium.com/@mortitech/sharedflow-vs-stateflow-a-comprehensive-guide-to-kotlin-flows-503576b4de31

```
Here are some best practices for using SharedFlow and StateFlow in Kotlin:

Choose the right flow:
Use SharedFlow when you need to broadcast values to multiple collectors or when you want to have multiple subscribers to the same stream of data.
Use StateFlow when you need to maintain and share a single source of truth for a state and automatically update all collectors with the latest state.

```

In this case, we only have one listener and single source.

forgerock-auth/src/main/java/org/forgerock/android/auth/AppAuthFragment2.kt
```diff
@@ -0,0 +1,93 @@
...
+import androidx.fragment.app.FragmentManager
+import kotlinx.coroutines.flow.MutableStateFlow
+import net.openid.appauth.AuthorizationResponse
+import org.forgerock.android.auth.centralize.BrowserLauncher
+import org.forgerock.android.auth.centralize.Launcher
+
+/**
+ * Headless Fragment to receive callback result from AppAuth library
+ */
+@RestrictTo(RestrictTo.Scope.LIBRARY)
```
jeyanthanperiyasamy, line 23: whats the issue with other approach ? i understand we sont need a extra call..but do you notice any issues on the extension approach ?

forgerock-auth/src/main/java/org/forgerock/android/auth/AppAuthFragment2.kt
```diff
@@ -0,0 +1,93 @@
...
+    companion object {
+        val TAG: String = AppAuthFragment2::class.java.name
+
+        /**
+         * Initialize the Fragment to receive AppAuth callback event.
+         */
+        @Synchronized
+        fun launch(activity: FragmentActivity, browser: FRUser.Browser) {
+            val fragmentManager: FragmentManager = activity.supportFragmentManager
+            var current = fragmentManager.findFragmentByTag(TAG) as AppAuthFragment2?
```
jeyanthanperiyasamy, line 58: var current = fragmentManager.findFragmentByTag(TAG) as? AppAuthFragment2

forgerock-auth/src/main/java/org/forgerock/android/auth/AppAuthFragment2.kt
```diff
@@ -0,0 +1,93 @@
...
+        BrowserLauncher.reset()
+    }
+
+    companion object {
+        val TAG: String = AppAuthFragment2::class.java.name
+
+        /**
+         * Initialize the Fragment to receive AppAuth callback event.
+         */
+        @Synchronized
```
jeyanthanperiyasamy, line 55: add

  @Synchronized
  @JVMstatic


Then the caller can remove the companion and call like this  AppAuthFragment2.launch

forgerock-auth/src/main/java/org/forgerock/android/auth/AppAuthFragment2.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        val state: MutableStateFlow<Intent?> = MutableStateFlow(null)
+        val delegate =
+            registerForActivityResult(AuthorizeContract()) {
+                state.value = it
+            }
+
+        val pending = savedInstanceState?.getBoolean("pending", false) ?: false
```
jeyanthanperiyasamy, line 34: just put this in val and reuse the key -  private val savedInstanceStateKey = "pending"

forgerock-auth/src/main/java/org/forgerock/android/auth/FRUser.java
```diff
@@ -317,35 +305,33 @@ private void login(Context context, FragmentManager manager, FRListener<FRUser>
...
                 return;
             }
 
-            this.listener = new FRListener<AuthorizationResponse>() {
+            this.listener = new FRListener<>() {
                 @Override
                 public void onSuccess(AuthorizationResponse result) {
                     InterceptorHandler interceptorHandler = InterceptorHandler.builder()
-                            .context(context)
+                            .context(activity.getApplicationContext())
```
jeyanthanperiyasamy, line 318: just   .context(activity)

samples/app/src/main/java/com/example/app/MainActivity.kt
```diff
@@ -11,10 +11,7 @@ import android.os.Bundle
 import androidx.activity.compose.setContent
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
-import androidx.compose.runtime.collectAsState
-import androidx.compose.runtime.getValue
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
-import org.forgerock.android.auth.FRUserKeys
 
 class MainActivity : AppCompatActivity() {
```
jeyanthanperiyasamy, line 16: For fully compose app which has only one ComponentActivity but does not have AppCompatActivity or FragmentActivity or fragments . is there a way to invoke this FRuser.browser.login() ?

forgerock-auth/src/main/java/org/forgerock/android/auth/AppAuthFragment2.kt
witrisna: Yes, the issue about "Don't keep activities" with consent. The issue is when finishing up browser authentication, the browser redirects back to the App, and the App calls onCreate, and resumes the previous state, when resumes previous state, it calls the launch again, which launches the browser again, which makes it never back to the App again.

forgerock-auth/src/main/java/org/forgerock/android/auth/AppAuthFragment2.kt
witrisna: Updated.

forgerock-auth/src/main/java/org/forgerock/android/auth/AppAuthFragment2.kt
witrisna: Updated.

forgerock-auth/src/main/java/org/forgerock/android/auth/AppAuthFragment2.kt
witrisna: Updated.

forgerock-auth/src/main/java/org/forgerock/android/auth/FRUser.java
witrisna: Updated.

jeyanthanperiyasamy (dismissed):

spetrov (approved):
👍🏻

PR #409, opened 2024-04-15
SDKS-3104 - Improvements in caching the key reference

jeyanthanperiyasamy:
# JIRA Ticket

Please link jira ticket here

# Description


forgerock-core/src/main/java/org/forgerock/android/auth/AndroidMEncryptor.kt
```diff
@@ -0,0 +1,85 @@
...
+    private val keyStore: KeyStore
+        get() {
+            val keyStore = KeyStore.getInstance(ANDROID_KEYSTORE)
+            keyStore.load(null)
+            return keyStore
+        }
+
+    @Throws(GeneralSecurityException::class, IOException::class)
+    override fun reset() {
+        keyStore.deleteEntry(keyAlias)
```
witrisna, line 78: Shoud we remove the cache also?

witrisna (changes requested):

forgerock-core/src/main/java/org/forgerock/android/auth/AndroidMEncryptor.kt
jeyanthanperiyasamy: yes good point

witrisna (approved):

rodrigoareis (dismissed):
Changes looks good to me

spetrov (dismissed):

spetrov (dismissed):

spetrov (approved):

PR #411, opened 2024-05-06
[SDKS-3201] Keep default constructor, the constructor is used with reflection, we want to keep the name when isMinifyEnabled is enabled.

witrisna:
# JIRA Ticket
[SDKS-3201](https://bugster.forgerock.org/jira/browse/SDKS-3201)

# Description

Keep default constructor, the constructor is used with reflection, we want to keep the name when isMinifyEnabled is enabled.


spetrov (approved):
Looks good to me!

jeyanthanperiyasamy (approved):

PR #413, opened 2024-05-15
SDKS-3199 Synchronize the encryption and decryption block to avoid keystore crash.

witrisna:
# JIRA Ticket

[SDKS-3199](https://bugster.forgerock.org/jira/browse/SDKS-3199)

# Description

Synchronize the encryption and decryption block to avoid keystore crash.


jeyanthanperiyasamy (approved):

spetrov (approved):
👍🏻

PR #415, opened 2024-05-15
SDKS-1710 WebAuthn Device Management

rodrigoareis:
# JIRA Ticket

[SDKS-1710](https://bugster.forgerock.org/jira/browse/SDKS-1710) WebAuthn Device Management

# Description

This change introduces a new method for removing WebAuthn keys from the server and locally in a single call.

Note: The ability to delete keys from the server relies on the availability of a new REST API for WebAuthn resources, which will be available on future versions of AM and IDC.


forgerock-auth/src/main/java/org/forgerock/android/auth/RemoteWebAuthnRepository.kt
```diff
@@ -0,0 +1,137 @@
...
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+
+private val TAG = RemoteWebAuthnRepository::class.java.simpleName
+
+internal class RemoteWebAuthnRepository(val serverConfig: ServerConfig = Config.getInstance().serverConfig) :
+    WebAuthnRepository {
+
+    override suspend fun delete(publicKeyCredentialSource: PublicKeyCredentialSource) {
+        val credentialId = String(publicKeyCredentialSource.id)
```
witrisna, line 34: Should it be base64?

forgerock-auth/src/main/java/org/forgerock/android/auth/RemoteWebAuthnRepository.kt
```diff
@@ -0,0 +1,137 @@
...
+        val userId = String(publicKeyCredentialSource.userHandle)
```
witrisna, line 35: Should it be base64?

forgerock-auth/src/main/java/org/forgerock/android/auth/RemoteWebAuthnRepository.kt
```diff
@@ -0,0 +1,137 @@
...
+
+        val findResponse = find(userId, credentialId)
+        val resourceId = findResponse.getString("uuid")
```
witrisna, line 38: getString will throw exception, may consider to use optString

forgerock-auth/src/main/java/org/forgerock/android/auth/RemoteWebAuthnRepository.kt
rodrigoareis: Looks like it's not required, the `toJson` and `fromJson` methods from the `PublicKeyCredentialSource` class handles the Base64 decode and decode.

forgerock-auth/src/main/java/org/forgerock/android/auth/RemoteWebAuthnRepository.kt
rodrigoareis: same as above

forgerock-auth/src/main/java/org/forgerock/android/auth/RemoteWebAuthnRepository.kt
rodrigoareis: good catch

witrisna (approved):

witrisna (dismissed):

forgerock-auth/src/main/java/org/forgerock/android/auth/webauthn/FRWebAuthn.kt
```diff
@@ -34,7 +50,26 @@ class FRWebAuthn(private val context: Context,
      * @param publicKeyCredentialSource The [PublicKeyCredentialSource] to be deleted
      */
     fun deleteCredentials(publicKeyCredentialSource: PublicKeyCredentialSource) {
-        webAuthnDataRepository.delete(publicKeyCredentialSource)
+        CoroutineScope(Dispatchers.IO).launch {
+            deleteCredentials(publicKeyCredentialSource, true)
```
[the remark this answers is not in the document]
spetrov: We've discussed this and decided to deprecate this method and preserve old behavior.

PR #420, opened 2024-06-04
SDKS-3022 OIDC ping discovery

jeyanthanperiyasamy:
# JIRA Ticket

Please link jira ticket here

# Description


forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -33,22 +45,80 @@ data class FROptions(val server: Server,
...
+            .build()
+
+        // Obtain instance of OkHttp client
+        val httpClient = getInstance().lookup(networkConfig)
+        return httpClient
+    }
+
+    suspend fun discover(discoverUrl: String): FROptions {
+        return withContext(Dispatchers.IO) {
+            suspendCancellableCoroutine { continuation ->
```
witrisna, line 78: I think we don't need to use suspendCancellableCoroutine, may just invoke httpClient.newCall().execute?

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -33,22 +45,80 @@ data class FROptions(val server: Server,
...
+                                response.code,
+                                response.message,
+                                response.body?.string()
+                            )
+                        }
+
+                        val openIdConfiguration = response.body?.string()?.let {
+                            JSONObject(it)
+                        }
+                        this@FROptions.urlPath.authorizeEndpoint =
```
witrisna, line 98: Can we keep FROptions immutable?
Clone the existing FROptions and replace the urlPath?
Something like this:
```kotlin
 suspend fun discover(discoverUrl: String) : FROptions {
        // Fetch the new UrlPath from the discoverUrl...
        val newUrlPath = UrlPath() // Replace this with the actual UrlPath fetched from the discoverUrl

        // Create a copy of the current FROptions object and modify the urlPath
        return this.copy(urlPath = newUrlPath)
    }
```

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -204,12 +274,12 @@ class SSLPinningBuilder {
...
  */
 data class UrlPath(val authenticateEndpoint: String? = null,
-                   val revokeEndpoint: String? = null,
+                   var revokeEndpoint: String? = null,
                    val sessionEndpoint: String? = null,
-                   val tokenEndpoint: String? = null,
-                   val userinfoEndpoint: String? = null,
-                   val authorizeEndpoint: String? = null,
-                   val endSessionEndpoint: String? = null)
+                   var tokenEndpoint: String? = null,
```
witrisna, line 279: Not sure if we can keep it immutable.

forgerock-auth/src/main/java/org/forgerock/android/auth/String+Extension.kt
```diff
@@ -0,0 +1,13 @@
+package org.forgerock.android.auth
+
+import java.net.URI
+import java.net.URISyntaxException
+
+ fun isAbsoluteUrl(url: String): Boolean {
+    return try {
+        val uri = URI(url)
+        uri.isAbsolute && uri.scheme != null && uri.host != null
```
witrisna, line 9: Calling isAbsolute is enough?
Does isAbsolute check the scheme?

samples/app/src/main/java/com/example/app/AppDrawer.kt
```diff
@@ -96,6 +97,16 @@ fun AppDrawer(
             },
             modifier = Modifier.padding(NavigationDrawerItemDefaults.ItemPadding)
         )
+        NavigationDrawerItem(
```
witrisna, line 100: Instead of adding a new navigation drawer, can we enhance com.example.app.env.EnvViewModel to support discover.

witrisna (changes requested):

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
```diff
@@ -34,21 +43,76 @@ data class FROptions(val server: Server,
...
     }
-    @Throws(IllegalArgumentException::class)
-    @JvmName("validateConfig")
-    internal fun validateConfig() {
-        require(server.url.isNotBlank()) { "AM URL cannot be blank" }
-        require(server.realm.isNotBlank()) { "Realm cannot be blank" }
-        require(server.cookieName.isNotBlank()) { "cookieName cannot be blank" }
+    private fun getOkHttpClient(url: URL): OkHttpClient {
+        val networkConfig = NetworkConfig.networkBuilder()
+            .timeUnit(TimeUnit.SECONDS)
```
witrisna, line 48: Why we need to set the timeUnit?

witrisna (changes requested):

forgerock-core/src/main/java/org/forgerock/android/auth/KotlinExtensions.kt
```diff
@@ -22,4 +24,15 @@ fun Long.convertToTime(pattern: String = "yyyyMMdd HH:mm:ss"): String {
     val date = Date(this)
     val simpleDateFormat = SimpleDateFormat(pattern, Locale.getDefault())
     return simpleDateFormat.format(date)
+}
+
+fun String?.isAbsoluteUrl(): Boolean {
+    return try {
+        this?.let {
+            val uri = URI(it)
+            uri.isAbsolute && uri.scheme != null && uri.host != null
```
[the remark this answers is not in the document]
jeyanthanperiyasamy: https://github.com/ForgeRock/forgerock-android-sdk/pull/420/files/c404174ba7b08a69b77c6f9bf8df3ae7142537f7#diff-84c9dad9fe7bf6d6d407d8388065f00d49e0d35d71e411084eeaf4522fdbbaeb

isabsolute checks the scheme as per the scheme...but it also checks the host ..See if this tests covers all the cases

forgerock-auth/src/main/java/org/forgerock/android/auth/FROptions.kt
jeyanthanperiyasamy: not required ,,,copt paste error

samples/app/src/main/java/com/example/app/centralize/CentralizeLoginViewModel.kt
```diff
@@ -10,35 +10,50 @@ package com.example.app.centralize
...
 import org.forgerock.android.auth.FRUser
 
 class CentralizeLoginViewModel : ViewModel() {
 
     var state = MutableStateFlow(CentralizeState())
         private set
+    fun login(fragmentActivity: FragmentActivity, options: FROptions) {
+        if(options.server.url.contains("pingone")) {
+            viewModelScope.launch {
+                val option = options.discover(options.server.url + "/.well-known/openid-configuration")
```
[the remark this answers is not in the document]
jeyanthanperiyasamy: yes i moved that ..i kind of want you see , what you think.

samples/app/src/main/java/com/example/app/env/EnvViewModel.kt
```diff
@@ -145,10 +159,20 @@ class EnvViewModel : ViewModel() {
         servers.add(local)
         servers.add(ops)
         servers.add(forgeblock)
+        servers.add(pingOIdc)
     }
 
     fun select(context: Context, options: FROptions) {
-        FRAuth.start(context, options)
+        if(options.server.url.contains("pingone")) {
```
witrisna, line 166: Using viewModelScope

samples/app/src/main/java/com/example/app/centralize/Centralize.kt
```diff
@@ -48,5 +50,4 @@ fun Centralize(centralizeLoginViewModel: CentralizeLoginViewModel) {
             Error(exception = this)
         }
     }
-}
-
```
witrisna, line 52: Should rollback Centralize  and CentraliszeLoginViewModel?

witrisna (changes requested):

witrisna (approved):

vahancouver (approved):
Changes look good. 

My only concern that we removed some validation along with their tests.
Is this actually necessary now that you have default Server value and setting the url from issuer field?!

jeyanthanperiyasamy:
My concern is if I keep the validation the realm and cookies have a default values now . But for ping users they can also override that and set that to empty. So it will blow the validation . We assume url will always be there with the issuer, but they can override that as well. So either you do only if all are mandatory or don’t do it

spetrov (dismissed):

samples/app/src/main/java/com/example/app/AppNavHost.kt
```diff
@@ -46,7 +46,8 @@ fun AppNavHost(navController: NavHostController,
         startDestination = startDestination) {
 
         composable(Destinations.ENV_ROUTE) {
-            val envViewModel = viewModel<EnvViewModel>()
+
+            val envViewModel: EnvViewModel = viewModel<EnvViewModel>()
```
rodrigoareis, line 50: Update copyright

forgerock-core/src/main/java/org/forgerock/android/auth/KotlinExtensions.kt
```diff
@@ -22,4 +24,15 @@ fun Long.convertToTime(pattern: String = "yyyyMMdd HH:mm:ss"): String {
     val date = Date(this)
     val simpleDateFormat = SimpleDateFormat(pattern, Locale.getDefault())
     return simpleDateFormat.format(date)
+}
+
+fun String?.isAbsoluteUrl(): Boolean {
```
rodrigoareis, line 29: Doc?

rodrigoareis (commented):
Overall changes looks good! Two minor comments to address

jeyanthanperiyasamy:
> Overall changes looks good! Two minor comments to address

Done

rodrigoareis (approved):
LGTM

witrisna (dismissed):

PR #427, opened 2024-06-19
SDKS-3282 FRUser.revokeAccessToken() should NOT end user's session

witrisna:
# JIRA Ticket

[SDKS-3282](https://bugster.forgerock.org/jira/browse/SDKS-3282)

# Description

- FRUser.revokeAccessToken() should NOT end user's session
- Browser logout should not invoke endSession endpoint directly

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/OAuth2Client.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/OAuth2Client.kt
@@ -8,6 +8,8 @@ package org.forgerock.android.auth
 
 import android.net.Uri
 import android.util.Base64
+import net.openid.appauth.AppAuthConfiguration
+import net.openid.appauth.EndSessionResponse
 import okhttp3.Call
 import okhttp3.Callback
 import okhttp3.FormBody
@@ -16,6 +18,8 @@ import okhttp3.Request
 import okhttp3.RequestBody
 import okhttp3.Response
 import org.forgerock.android.auth.Logger.Companion.debug
+import org.forgerock.android.auth.centralize.AppAuthFragment2.Companion.endSession
+import org.forgerock.android.auth.centralize.EndSessionInput
 import org.forgerock.android.auth.exception.AuthorizeException
 import java.io.IOException
 import java.io.UnsupportedEncodingException
@@ -154,7 +158,7 @@ class OAuth2Client(
      * @param accessToken The AccessToken to be revoked
      * @param listener    Listener to listen for revoke event
      */
-    fun revoke(accessToken: AccessToken, listener: FRListener<Void?>?) {
+    fun revoke(accessToken: AccessToken, listener: FRListener<Void?>) {
         revoke(accessToken, true, listener)
     }
 
@@ -166,7 +170,7 @@ class OAuth2Client(
      * @param useRefreshToken If true, revoke with refresh token, otherwise revoke access token
      * @param listener        Listener to listen for revoke event
      */
-    fun revoke(accessToken: AccessToken, useRefreshToken: Boolean, listener: FRListener<Void?>?) {
+    fun revoke(accessToken: AccessToken, useRefreshToken: Boolean, listener: FRListener<Void?>) {
         debug(TAG, "Revoking Access Token & Refresh Token")
         val handler = OAuth2ResponseHandler()
         try {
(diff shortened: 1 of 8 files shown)


samples/app/src/main/java/com/example/app/token/TokenViewModel.kt
spetrov, line 1: Update copyright note

samples/app/src/main/java/com/example/app/token/Token.kt
spetrov, line 1: Update copyright note

samples/app/src/main/java/com/example/app/token/Token.kt
```diff
@@ -43,11 +43,15 @@ fun Token(tokenViewModel: TokenViewModel) {
             horizontalArrangement = Arrangement.SpaceEvenly) {
             Button(
                 onClick = { tokenViewModel.getAccessToken() }) {
-                Text(text = "Refresh")
+                Text(text = "Get")
             }
             Button(
                 onClick = { tokenViewModel.forceRefresh() }) {
-                Text(text = "ForceRefresh")
```
spetrov, line 50: `ForceRefresh` button was good for testing. I see that now the 4 buttons don't fit in the UI, but can we possibly introduce a checkbox for the force refresh?

spetrov (changes requested):

samples/app/src/main/java/com/example/app/token/TokenViewModel.kt
witrisna: Updated

samples/app/src/main/java/com/example/app/token/Token.kt
witrisna: Updated

samples/app/src/main/java/com/example/app/token/Token.kt
witrisna: CleanShot 2024-06-19 at 10 06 50@2x

Using short form, hope this fix all screen

forgerock-auth/src/main/java/org/forgerock/android/auth/OAuth2Client.kt
```diff
@@ -154,7 +158,7 @@ class OAuth2Client(
      * @param accessToken The AccessToken to be revoked
      * @param listener    Listener to listen for revoke event
      */
-    fun revoke(accessToken: AccessToken, listener: FRListener<Void?>?) {
+    fun revoke(accessToken: AccessToken, listener: FRListener<Void?>) {
```
rodrigoareis, line 161: Would this change to the second parameter affect the backward compatibility?

forgerock-auth/src/main/java/org/forgerock/android/auth/TokenManager.java
```diff
@@ -88,5 +91,13 @@ public interface TokenManager {
      */
     void revoke(FRListener<Void> listener);
 
+    /**
+     * OAuth2 Token Revocation, and end the session
+     *
+     * @param appAuthConfiguration The AppAuthConfiguration
+     * @param listener             Listener to listen for token revocation event.
+     */
+    void revokeAndEndSession(Supplier<AppAuthConfiguration> appAuthConfiguration, FRListener<Void> listener);
```
rodrigoareis, line 100: Should we make second parameter optional, to avoid the need to pass null?

samples/app/src/main/java/com/example/app/env/EnvViewModel.kt
```diff
@@ -46,9 +44,9 @@ class EnvViewModel : ViewModel() {
 
     val dbind = FROptionsBuilder.build {
         server {
-            url = "https://openam-updbind.forgeblocks.com/am"
-            realm = "bravo"
-            cookieName = "ccdd0582e7262db"
+            url = "http://192.168.86.32:8080/openam"
```
rodrigoareis, line 47: Do we want to commit this change?

rodrigoareis (commented):
Changes looks good to me. Just minor questions to be addressed.

forgerock-auth/src/main/java/org/forgerock/android/auth/OAuth2Client.kt
witrisna: The interface is for internal use only, I don't think any developer is using this interface. And I have confirmed that all caller in the SDK pass a non null object.

forgerock-auth/src/main/java/org/forgerock/android/auth/TokenManager.java
witrisna: For now the SessionManager.close() ignores the result of success or failure, tries best to close the session.
Would leave it to take a listener for now, we may want to receive result in the future.

samples/app/src/main/java/com/example/app/env/EnvViewModel.kt
witrisna: Yes, this file is more for testing.
Environment keep changing.
We may eventually use a valid environment, so that developer can test with our environment, and add this to .ignore file.
We can discuss about the approach later, but for now, this is a file for internal testing.

rodrigoareis (approved):
Changes looks good to me

spetrov (approved):

PR #428, opened 2024-06-19
SDKS-3284 Inconsistent behavior when forgerock_url is not provided.

witrisna:
# JIRA Ticket

[SDKS-3284](https://bugster.forgerock.org/jira/browse/SDKS-3284)

# Description

- Validate the server url before persist, for now, it does not validate the format, just make sure it is not empty.
- Ignore exception when failed to load the stored configuration.


rodrigoareis (dismissed):
Changes looks good to me

spetrov (approved):
Looks good to me!

PR #429, opened 2024-06-20
SDKS-3283 Persisting the signOutRedirectUri into the shared preferences.

spetrov:
# JIRA Ticket

[SDKS-3283](https://bugster.forgerock.org/jira/browse/SDKS-3283) Starting the SDK with different configuration does NOT end user's session 

# Description

- Updated the `ConfigHelper` class to store the `signOutRedirectUri` into the share preferences
- Added back the getters to the path config variables (`authenticateEndpoint`, `authorizeEndpoint`, `tokenEndpoint`, `revokeEndpoint`, `userinfoEndpoint`, `sessionEndpoint`, `endSessionEndpoint`) to align the behavior with iOS

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
@@ -54,19 +54,12 @@ public class Config {
     private CookieJar cookieJar;
     private String cookieName;
 
-    @Getter(value = AccessLevel.NONE)
     private String authenticateEndpoint;
-    @Getter(value = AccessLevel.NONE)
     private String authorizeEndpoint;
-    @Getter(value = AccessLevel.NONE)
     private String tokenEndpoint;
-    @Getter(value = AccessLevel.NONE)
     private String revokeEndpoint;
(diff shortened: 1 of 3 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
```diff
@@ -54,19 +54,12 @@ public class Config {
     private CookieJar cookieJar;
     private String cookieName;
 
-    @Getter(value = AccessLevel.NONE)
```
jeyanthanperiyasamy, line 57: is this committed accidentally ? can you check this  @spetrov

jeyanthanperiyasamy (changes requested):
please look comments

forgerock-auth/src/main/java/org/forgerock/android/auth/Config.java
jeyanthanperiyasamy: i think i resolved this comment with stoyan

jeyanthanperiyasamy (approved):

rodrigoareis (approved):
Looks good to me

PR #434, opened 2024-07-22
SDKS-3227_Skipping type 4 TextOutputCallbacks

george-bafaloukas-forgerock:
Added conditional sbmittion of TextOutput Callbacks based on the type skipping 4

Updated the sample apps to handle the ScriptedTextOutputCallbacks and long content better

# JIRA Ticket

[Please link jira ticket here](https://bugster.forgerock.org/jira/browse/SDKS-3227)

# Description

Added conditional sbmittion of TextOutput Callbacks based on the type skipping 4

Updated the sample apps to handle the ScriptedTextOutputCallbacks and long content better

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/Node.java
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/Node.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019 - 2023 ForgeRock. All rights reserved.
+ * Copyright (c) 2019 - 2024 ForgeRock. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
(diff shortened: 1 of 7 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/Node.java
```diff
@@ -59,7 +62,16 @@ JSONObject toJsonObject() throws JSONException {
         }
         JSONArray array = new JSONArray();
         for (Callback cb : callbacks) {
-            array.put(new JSONObject(cb.getContent()));
+            if ((cb instanceof TextOutputCallback)) {
```
witrisna, line 65: I would handle it on receiving the Callback, 
org.forgerock.android.auth.callback.TextOutputCallback#setAttribute
Removing the content when msg type is 4

witrisna (changes requested):

forgerock-auth/src/main/java/org/forgerock/android/auth/Node.java
george-bafaloukas-forgerock: Receiving the callback is no problem for the SDK or the server side. Submitting back is what causes issues. I am of the opinion that if we keep the callback and just remove it before submitting, we are more open to anyone using this to pass data (valuable ones not a JS script) back to the SDK. So instead of dropping the support, we just make that an non-submittable callback. Let's discuss this when you come online.

spetrov (dismissed):
Looks good to me!

witrisna (dismissed):

spetrov (approved):
Approved!

PR #435, opened 2024-08-06
SDKS-3340 Added PingAM, PingAIC, PingOne and PingFederate examples

forgerock-chris:
- Added (fake) example settings for all the server types we currently support, using the same/similar values as the docs.
- Altered the select function to handle "raw" .well-known URIs. I put this first in case they use a PingOne `.well-known` value, as currently the code would append another `/.well-known/openid-configuration` string if it sees "pingone" in the URL.

Had an issue where the app would crash when attempting to select an env that didn't return an HTTP status code (i.e DNS didn't resolve).

To try and fix that I've surrounded the "discover" calls with a try...catch, but I don't really know what I'm doing so I think that could use some focus please.


jeyanthanperiyasamy (commented):
Looks good. but we need to agree as a team , is this make sense

spetrov:
This was done in #451

PR #436, opened 2024-08-07
SDKS-3297 Support MetadataCallback used by Marketplace PingOne Protect

rodrigoareis:
# JIRA Ticket

[SDKS-3297](https://bugster.forgerock.org/jira/browse/SDKS-3297) [Android] Support Marketplace PingOne Protect nodes

# Description

Make sure that the Android SDK supports the PingOne Protect nodes offered in the Marketplace, which utilized the MetadataCallback.

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/AbstractProtectCallback.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/AbstractProtectCallback.kt
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) 2024 ForgeRock. All rights reserved.
+ *
+ *  This software may be modified and distributed under the terms
+ *  of the MIT license. See the LICENSE file for details.
+ */
+
+package org.forgerock.android.auth.callback
+
+import org.forgerock.android.auth.Node
+import org.json.JSONException
+import org.json.JSONObject
+
+const val _ACTION = "_action"
+const val _TYPE = "_type"
+const val CLIENT_ERROR = "clientError"
+const val PING_ONE_RISK_EVALUATION_SIGNALS = "pingone_risk_evaluation_signals"
+const val PING_ONE_PROTECT = "PingOneProtect"
+const val PROTECT_INITIALIZE = "protect_initialize"
+const val PROTECT_RISK_EVALUATION = "protect_risk_evaluation"
+const val PING_ONE_PROTECT_INITIALIZE_CALLBACK = "PingOneProtectInitializeCallback"
+const val PING_ONE_PROTECT_EVALUATION_CALLBACK = "PingOneProtectEvaluationCallback"
+
+/**
+ * Abstract Protect Callback that provides the raw content of the Callback, and common methods
+ * for sub classes to access.
+ */
+abstract class AbstractProtectCallback(raw: JSONObject, index: Int) : NodeAware, AbstractCallback(raw, index) {
+
+    /**
+     * Constructor for [AbstractProtectCallback]. Capable of parsing the [MetadataCallback] used for Protect.
+     */
+    init {
+        val type = raw.optString("type")
+        if (type == "MetadataCallback") {
+            val output = raw.optJSONArray("output")
+            if (output != null) {
+                val nameValuePair = output.getJSONObject(0)
+                val name = nameValuePair.optString("name")
+                if (nameValuePair != null && name == "data") {
+                    val value = nameValuePair.getJSONObject("value")
+                    for (attr in value.keys()) {
+                        setAttribute(attr, value.get(attr))
+                    }
+                }
+            }
+        }
+    }
+
+    /**
+     * The [Node] that associate with this Callback
+     */
+    private lateinit var node: Node
+
+    override fun setNode(node: Node) {
+        this.node = node
+    }
+
+    /**
+     * Set the client error to the [HiddenValueCallback] which associated with the Protect
+     * Callback.
+     *
+     * @param value The Value to set to the [HiddenValueCallback]
+     */
+    fun setClientErrorInHiddenCallback(value: String) {
+        for (callback in node.callbacks) {
+            if (callback is HiddenValueCallback) {
+                if (callback.id == CLIENT_ERROR) {
+                    callback.value = value
+                }
+            }
+        }
+    }
+
+    /**
+     * Set the signals to the [HiddenValueCallback] which associated with the callback.
+     *
+     * @param value The Value to set to the [HiddenValueCallback].
+     */
+    fun setSignalsInHiddenCallback(value: String) {
+        for (callback in node.callbacks) {
+            if (callback is HiddenValueCallback) {
+                if (callback.id == PING_ONE_RISK_EVALUATION_SIGNALS) {
+                    callback.value = value
+                }
+            }
+        }
+    }
+
+    companion object {
+
+        /**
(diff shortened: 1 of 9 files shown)


forgerock-auth/src/main/java/org/forgerock/android/auth/callback/AbstractProtectCallback.kt
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Abstract Protect Callback that provides the raw content of the Callback, and common methods
+ * for sub classes to access.
+ */
+abstract class AbstractProtectCallback(raw: JSONObject, index: Int) : NodeAware, AbstractCallback(raw, index) {
+
+    /**
+     * Constructor for [AbstractProtectCallback]. Capable of parsing the [MetadataCallback] used for Protect.
+     */
+    init {
```
jeyanthanperiyasamy, line 33: we can refactor something like this..not a big deal


init {
    val type = raw.optString("type")
    if (type == "MetadataCallback") {
        raw.optJSONArray("output")?.let { output ->
            output.getJSONObject(0)?.let { nameValuePair ->
                if (nameValuePair.optString("name") == "data") {
                    nameValuePair.optJSONObject("value")?.let { value ->
                        value.keys().forEach { attr ->
                            setAttribute(attr, value.get(attr))
                        }
                    }
                }
            }
        }
    }
}

jeyanthanperiyasamy (commented):
Adding few more comments

ping-protect/src/main/java/org/forgerock/android/auth/PingOneProtectInitializeCallback.kt
```diff
@@ -19,12 +19,8 @@ private val TAG = PingOneProtectInitializeCallback::class.java.simpleName
 /**
  * Callback to initialize the ping one protect
  */
-open class PingOneProtectInitializeCallback : AbstractCallback {
-    @Keep
-    constructor(jsonObject: JSONObject, index: Int) : super(jsonObject, index)
-
-    @Keep
-    constructor() : super()
```
jeyanthanperiyasamy, line 27: i am not sure this will fail , if we minify the project, we may need both constructors, can we just do this PingOneProtectInitializeCallback: AbstractProtectCallback

ping-protect/src/main/java/org/forgerock/android/auth/PingOneProtectEvaluationCallback.kt
```diff
@@ -8,21 +8,19 @@ package org.forgerock.android.auth
...
 private val TAG = PingOneProtectEvaluationCallback::class.java.simpleName
 
 /**
  * Callback to evaluate Ping One Protect
  */
-open class PingOneProtectEvaluationCallback : AbstractCallback {
+open class PingOneProtectEvaluationCallback @Keep constructor(jsonObject: JSONObject, index: Int) :
+    AbstractProtectCallback(jsonObject, index) {
 
-    @Keep
```
jeyanthanperiyasamy, line 21: same here. lets keep both PingOneProtectEvaluationCallback: AbstractProtectCallback

forgerock-auth/src/main/java/org/forgerock/android/auth/callback/AbstractProtectCallback.kt
```diff
@@ -0,0 +1,128 @@
...
+const val PROTECT_RISK_EVALUATION = "protect_risk_evaluation"
+const val PING_ONE_PROTECT_INITIALIZE_CALLBACK = "PingOneProtectInitializeCallback"
+const val PING_ONE_PROTECT_EVALUATION_CALLBACK = "PingOneProtectEvaluationCallback"
+
+/**
+ * Abstract Protect Callback that provides the raw content of the Callback, and common methods
+ * for sub classes to access.
+ */
+abstract class AbstractProtectCallback(raw: JSONObject, index: Int) : NodeAware, AbstractCallback(raw, index) {
+
```
jeyanthanperiyasamy, line 29: Here we can keep two constructors

@Keep
    constructor(jsonObject: JSONObject, index: Int) : super(jsonObject, index)

    @Keep
    constructor() : super()

jeyanthanperiyasamy (commented):
overall looks good , just few comments need to be addressed

ping-protect/src/main/java/org/forgerock/android/auth/PingOneProtectEvaluationCallback.kt
```diff
@@ -8,21 +8,19 @@ package org.forgerock.android.auth
...
+import org.forgerock.android.auth.callback.PING_ONE_RISK_EVALUATION_SIGNALS
 import org.json.JSONObject
 
 private val TAG = PingOneProtectEvaluationCallback::class.java.simpleName
 
 /**
  * Callback to evaluate Ping One Protect
  */
-open class PingOneProtectEvaluationCallback : AbstractCallback {
+open class PingOneProtectEvaluationCallback @Keep constructor(jsonObject: JSONObject, index: Int) :
```
witrisna, line 21: We have to use @keep for both constructors, currently, we use reflection to construct the class. You may want to enable 
`    isMinifyEnabled = true
    isShrinkResources = true`
in the AndroidBuildGradlePlugin, build a release build, install the apk.

ping-protect/src/main/java/org/forgerock/android/auth/PingOneProtectEvaluationCallback.kt
```diff
@@ -8,21 +8,19 @@ package org.forgerock.android.auth
 
 import android.content.Context
 import androidx.annotation.Keep
-import org.forgerock.android.auth.callback.AbstractCallback
+import org.forgerock.android.auth.callback.AbstractProtectCallback
```
witrisna, line 11: Remove unnecessary import

ping-protect/src/main/java/org/forgerock/android/auth/PingOneProtectEvaluationCallback.kt
```diff
@@ -44,6 +42,7 @@ open class PingOneProtectEvaluationCallback : AbstractCallback {
      */
     fun setSignals(value: String) {
         super.setValue(value, 0)
```
witrisna, line 44: Any side effect for setting this for marketplace node? There should be no input for marketplace node.

witrisna (changes requested):

ping-protect/src/main/java/org/forgerock/android/auth/PingOneProtectEvaluationCallback.kt
rodrigoareis: done as part of Jey's comments

ping-protect/src/main/java/org/forgerock/android/auth/PingOneProtectEvaluationCallback.kt
rodrigoareis: good catch!

ping-protect/src/main/java/org/forgerock/android/auth/PingOneProtectEvaluationCallback.kt
rodrigoareis: you mean AbstractProtectCallback? we need this import

spetrov (approved):
Looks good to me! 👍🏻

witrisna (approved):

jeyanthanperiyasamy (approved):

PR #448, opened 2024-09-13
SDKS-3426 When compare the SessionToken associated with the AccessToken, it should just compare the value but not the SSOToken Object.

witrisna:
# JIRA Ticket

[SDKS-3426](https://bugster.forgerock.org/jira/browse/SDKS-3426)

# Description

SDKS-3426 When compare the SessionToken associated with the AccessToken, it should just compare the value but not the SSOToken Object.


spetrov (approved):
Looks right to me!

PR #451, opened 2024-09-23
SDKS-3420 Use sample environment for Sample App.

witrisna:
# JIRA Ticket

[SDKS-3420](https://bugster.forgerock.org/jira/browse/SDKS-3420)

# Description

SDKS-3420 Use sample environment for Sample App.

### DIFF
--- a/samples/app/src/main/java/com/example/app/env/EnvViewModel.kt
+++ b/samples/app/src/main/java/com/example/app/env/EnvViewModel.kt
@@ -1,8 +1,8 @@
 /*
- * Copyright (c) 2023 - 2024 ForgeRock. All rights reserved.
+ * Copyright (c) 2023-2024 ForgeRock. All rights reserved.
  *
- * This software may be modified and distributed under the terms
- * of the MIT license. See the LICENSE file for details.
+ *  This software may be modified and distributed under the terms
+ *  of the MIT license. See the LICENSE file for details.
  */
 
 package com.example.app.env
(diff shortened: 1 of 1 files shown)


samples/app/src/main/java/com/example/app/env/EnvViewModel.kt
```diff
@@ -13,59 +13,107 @@ import androidx.compose.runtime.mutableStateOf
...
+        servers.add(PingAM)
+        servers.add(PingAdvancedIdentityCloud)
+        servers.add(PingOne)
+        servers.add(PingFederate)
     }
 
     fun select(context: Context, options: FROptions) {
-        if(options.server.url.contains("pingone")) {
+        // Discover settings from the specified PingOne server's .well-known endpoint
+        if(options.server.url.contains("pingone") || options.server.url.contains("pingfederate")) {
```
forgerock-chris, line 104: As PingFed is self-hosted, the chance of the domain containing "pingfederate" is very low.

In addition to the "pingone" check, could we have a separate `if` to check for `.well-known/openid-configuration`, and use `discover` if found?

forgerock-chris (dismissed):
Thanks for adopting these changes over Andy. I'll update my tutorials to make the steps for editing the sample a lot simpler.

samples/app/src/main/java/com/example/app/env/EnvViewModel.kt
forgerock-chris: To clarify a bit, in the instructions I'll say something along the lines of:
-  Enter the .well-known URI of your PingFederate server.
If that URL _does not_ contain the word "pingfederate", the sample won't work "out of the box", and I'll need an additional instruction to make sure that "discover" is used.
Hope that makes sense?

spetrov (dismissed):

samples/app/src/main/java/com/example/app/env/EnvViewModel.kt
witrisna: I have used the instance instead:
```
if(options === PingOne || options === PingFederate) {
```

forgerock-chris: Okay, I'll update the PingFederate instructions to match, cheers Andy.
I see this is heading for `develop`, will it also go to `main`, or only when there's a new release?

witrisna: Only when there is a new release.

forgerock-chris (approved):

spetrov (approved):

PR #452, opened 2024-09-23
SDKS-3373 - ReCaptcha e2e test cases

spetrov:
# JIRA Ticket

[SDKS-3373](https://bugster.forgerock.org/jira/browse/SDKS-3373) [QA] Test end-to-end flow of ReCaptcha Enterprise

# Description

- Added e2e test cases for the ReCaptcha Enterprise feature. These test cases test both server and client side. These require _live_ server and the following authentication tree: 
image


forgerock-integration-tests/build.gradle.kts
```diff
@@ -43,11 +43,17 @@ dependencies {
     androidTestImplementation(libs.androidx.biometric.ktx)
     androidTestImplementation(libs.nimbus.jose.jwt)
     androidTestImplementation(libs.okhttp)
+    androidTestImplementation(libs.kotlinx.coroutines.test)
+    testImplementation(libs.kotlinx.coroutines.test)
```
jeyanthanperiyasamy, line 47: do we need this testImplementation(libs.kotlinx.coroutines.test) ?,what if you remove this and run the test

forgerock-auth/src/main/java/org/forgerock/android/auth/callback/ReCaptchaEnterpriseCallback.kt
```diff
@@ -85,6 +83,15 @@ class ReCaptchaEnterpriseCallback : AbstractCallback {
         super.setValue(value.toString(), 3)
     }
 
+    /**
+     * Set the Action for the ReCAPTCHA
+     *
+     * @param value The Action
+     */
+    private fun setAction(value: String) {
```
jeyanthanperiyasamy, line 91: i agree with his change, we dont want to let the user to set the action

forgerock-integration-tests/build.gradle.kts
spetrov: Yeah, looks like this is not needed indeed - I removed it.

jeyanthanperiyasamy (approved):

witrisna (approved):

PR #460, opened 2024-10-17
SDKS-3489 Upgrade to Gradle 8.9 and Support Android Studio Ladybug

witrisna:
# JIRA Ticket

[SDKS-3489](https://pingidentity.atlassian.net/browse/SDKS-3489)

# Description

SDKS-3489 Upgrade to Gradle 8.9 and Support Android Studio Ladybug


jeyanthanperiyasamy (approved):

jeyanthanperiyasamy (approved):
Just a question , what happens if someone does not have java 17 installed ? if they have only java 11 or 21 installed , would this change works ?

witrisna:
> Just a question , what happens if someone does not have java 17 installed ? if they have only java 11 or 21 installed , would this change works ?

I am not sure, I believe it will failed with error msg.
The change is to make the build consistent, will always compile with 17, but not depends on the IDE or project setting.

spetrov (approved):
👍🏻

PR #464, opened 2024-11-20
SDKS-3595 Remove Deprecated Sample Apps

witrisna:
# JIRA Ticket

[SDKS-3593](https://pingidentity.atlassian.net/browse/SDKS-3593)

# Description

SDKS-3595 Remove Deprecated Sample Apps


e2e/app/src/main/java/com/example/app/env/EnvViewModel.kt
```diff
@@ -0,0 +1,213 @@
...
+
+const val USER_PROFILE_JOURNEY = "UserProfile"
+
+class EnvViewModel : ViewModel() {
+
+    val servers = mutableListOf<FROptions>()
+
+    val localhost = FROptionsBuilder.build {
+        server {
+            url = "https://andy.petrov.ca/openam"
```
spetrov, line 33: Not sure if we should keep these environments in the sample app... 
I think our docs has some references to this file... Let's confirm with @forgerock-chris though...

spetrov (approved):
LGTM, but let's confirm if we want to keep all the environments in the sample app...

e2e/app/src/main/java/com/example/app/env/EnvViewModel.kt
rodrigoareis: I agree, we should remove those. Ideally we should load these configs from a file, and ignore this file on commits.

rodrigoareis (dismissed):
Looks good to me.

e2e/app/src/main/java/com/example/app/env/EnvViewModel.kt
witrisna: Revert to sample env.

spetrov (approved):
LGTM! 👍🏻

PR #469, opened 2025-02-03
SDKS-3772 Revoke and terminate the session when it does not match the existing session.

witrisna:
# JIRA Ticket
[SDKS-3772](https://pingidentity.atlassian.net/browse/SDKS-3772)

# Description

SDKS-3772 Revoke and terminate the session when it does not match the existing session.


spetrov (approved):
LGTM! 👍🏻

PR #471, opened 2025-02-21
SDKS-3684 Handle push device token updates

rodrigoareis:
# JIRA Ticket

[SDKS-3684](https://pingidentity.atlassian.net/browse/SDKS-3684) Handle push device token updates

# Description

To accommodate modifications made to the AM Push Notification Service for device token updates, client-side changes are necessary.  A method within the FRAClient class should be created to manage the receipt of a device token from Firebase Cloud Messaging (FCM) and subsequently transmit that token to the server for storage or updating.

### DIFF
--- a/forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
+++ b/forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
@@ -21,14 +21,13 @@ import org.forgerock.android.auth.exception.InvalidNotificationException;
 import org.forgerock.android.auth.exception.MechanismCreationException;
 import org.forgerock.android.auth.exception.MechanismParsingException;
 import org.forgerock.android.auth.exception.MechanismPolicyViolationException;
+import org.forgerock.android.auth.exception.MechanismUpdatePushTokenException;
 import org.forgerock.android.auth.policy.FRAPolicy;
 
 import java.util.ArrayList;
-import java.util.Calendar;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.TimeZone;
 import java.util.concurrent.atomic.AtomicInteger;
 
 class AuthenticatorManager {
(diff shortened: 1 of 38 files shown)


spetrov (approved):
Looks good to me!

forgerock-authenticator/src/main/java/org/forgerock/android/auth/MechanismFactory.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 - 2023 ForgeRock. All rights reserved.
```
witrisna, line 2: Looks like no change on this file.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
```diff
@@ -251,25 +255,78 @@ PushNotification getNotificationByMessageId(String messageId) {
...
-                        "device token (%s) was received.", this.deviceToken, newDeviceToken);
-                throw new AuthenticatorException("The SDK was initialized with a different deviceToken.");
+                        "device token (%s) was received.", this.pushDeviceTokenManager.getDeviceTokenId(), deviceToken);
+                throw new AuthenticatorException("The SDK was initialized with a different deviceToken. " +
+                        "Use FRAClient#updateDeviceToken method to update the device token.");
             }
         }
     }
 
+    void updateDeviceToken(String newDeviceToken, FRAListener<Boolean> listener) {
```
witrisna, line 275: I think it makes sense to provide this convenient method for developer, most of the time they want to update all the mechanisms, however we may want to improve the signature. If reusing FRAListener, we may want to use FRAListener<Void>, throw exception when one or more mechanism is failed. You may want to reuse AuthenticatorException.

It is kind of wierd that the developer implement onException() but it will never be called even failed.

rodrigoareis: I agree, I have updated the code following your suggestion.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/PushFactory.java
```diff
@@ -43,19 +42,20 @@ class PushFactory extends MechanismFactory {
      *
      * @param context The application context
      * @param storageClient The storage system
-     * @param fcmToken The FCM device registration token an Android app needs to receive push messages
+     * @param pushDeviceTokenManager Manages the FCM device registration token an Android app needs to receive push messages
      */
-    public PushFactory(Context context, StorageClient storageClient, String fcmToken) {
+    public PushFactory(Context context, StorageClient storageClient, PushDeviceTokenManager pushDeviceTokenManager) {
```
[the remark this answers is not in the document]
rodrigoareis: This class isn't public, so we are good.

forgerock-authenticator/src/main/java/org/forgerock/android/auth/MechanismFactory.java
rodrigoareis: Reverted

forgerock-authenticator/src/main/java/org/forgerock/android/auth/AuthenticatorManager.java
```diff
@@ -251,25 +254,81 @@ PushNotification getNotificationByMessageId(String messageId) {
...
+                @Override
+                public void onException(Exception e) {
+                    Logger.warn(TAG, "Error updating FCM device token for mechanism %s.", pushMechanism.getMechanismUID(), e);
+                    failedMechanisms.add(pushMechanism);
+                    if (completedMechanisms.incrementAndGet() == totalMechanisms) {
+                        // All mechanisms have completed, but some failed
+                        listener.onException(new MechanismUpdatePushTokenException(
+                                "Error updating FCM device token for some mechanisms.", failedMechanisms));
+                    }
+                }
```
witrisna, line 317: Call listener.onException(e)? otherwise the caller will not receive anything.

rodrigoareis: @witrisna, we are calling it just when all mechanisms are complete. See that we also call `listener.onException()` in `onSucess()`

witrisna (dismissed):

spetrov (dismissed):
LGTM!

spetrov (approved):

spetrov (approved):
LGTM!

PR #474, opened 2025-04-15
SDKS-3771 Setup Mend SAST scan

spetrov:
# JIRA Ticket

[SDKS-3771](https://pingidentity.atlassian.net/browse/SDKS-3771) Setup Mends SAST scan

# Description

Setup Mend SAST scan. 
A few notes:
- Currently, the pipeline will fail if some **Critical** or **High** vulnerabilities are found! Mediums and Lows won't fail the pipeline! We can change that easily if needed...
- Currently the `.build` directory is excluded from the scans. This is controlled by the `MEND_SAST_PATH_EXCLUSIONS` env variable. To modify the list we can simply change the github action variable `MEND_SAST_PATH_EXCLUSIONS` [here](https://github.com/ForgeRock/ping-ios-sdk/settings/variables/actions)...
- The pipeline attaches/publishes the SAST scan results as an attachment to the pipeline execution run
- The results are pushed to our Mend project - direct URL is available through the attached results and also in the slack message:
image

### DIFF
--- a/.github/workflows/mend-cli-scan.yaml
+++ b/.github/workflows/mend-cli-scan.yaml
@@ -25,14 +25,6 @@ jobs:
           repository: ${{github.event.pull_request.head.repo.full_name}}
           fetch-depth: 0
 
-      # Setup JDK and cache and restore dependencies.
-      - name: Set up JDK 17
-        uses: actions/setup-java@v3
-        with:
-          distribution: 'temurin'
-          java-version: '17'
-          cache: 'gradle'
-
       # Setup Mend CLI
       - name: Download and cache the Mend CLI executable
         id: cache-mend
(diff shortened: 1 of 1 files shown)


.github/workflows/mend-cli-scan.yaml
```diff
@@ -25,14 +25,6 @@ jobs:
           repository: ${{github.event.pull_request.head.repo.full_name}}
           fetch-depth: 0
 
-      # Setup JDK and cache and restore dependencies.
```
spetrov, line 28: There is no need to setup JDK on the runner for Mend CLI scans...

.github/workflows/mend-cli-scan.yaml
```diff
@@ -49,45 +41,118 @@ jobs:
       - if: ${{ steps.cache-mend.outputs.cache-hit != 'true' }}
         name: Download Mend CLI executable (cache miss...)
         continue-on-error: true
+        shell: bash
```
spetrov, line 44: Even though GitHub defaults to `bash` on Ubuntu runners, it's safer to be explicit for portability...

.github/workflows/mend-cli-scan.yaml
```diff
@@ -49,45 +41,118 @@ jobs:
...
+          echo "MEND_SCA_MEDIUM_COUNT=$MEND_SCA_MEDIUM_COUNT" >> $GITHUB_ENV
+          echo "MEND_SCA_LOW_COUNT=$MEND_SCA_LOW_COUNT" >> $GITHUB_ENV
 
-      # Check for failures and set the outcome of the workflow
-      - name: Parse the result and set job status
+      # Check for failures in SCA scan and set the outcome of the workflow
+      - name: Fail if Critical or High SCA vulnerabilities are found
+        shell: bash
+        run: |
+          if [ "$MEND_SCA_CRITICAL_COUNT" -gt 0 ] || [ "$MEND_SCA_HIGH_COUNT" -gt 0 ]; then
```
spetrov, line 79: Note that the job fails only if some **Critical** or **High** vulnerabilities are found! Mediums and Lows won't fail the pipeline!

.github/workflows/mend-cli-scan.yaml
```diff
@@ -49,45 +41,118 @@ jobs:
...
+      #       Execute Mend SAST scan         #
+      ########################################
+      - name: Mend SAST scan
+        env:
+          MEND_EMAIL: ${{ secrets.MEND_EMAIL }}
+          MEND_USER_KEY: ${{ secrets.MEND_USER_KEY }}
+          MEND_URL: ${{ vars.MEND_SERVER_URL }}
+        shell: bash
+        run: |
+          mend code --report --filename ${{ vars.MEND_SAST_REPORT_NAME }} --formats json,pdf --non-interactive --scope ${{ vars.MEND_PRODUCT_NAME }}//${{ vars.MEND_PROJECT_NAME }} > mend-sast-scan-result.txt
```
spetrov, line 108: We run the mend SAST scan and produce 2 files - json and pdf + the report is published to Mend under ${{ vars.MEND_PRODUCT_NAME }}//${{ vars.MEND_PROJECT_NAME }}
The json file is used to parse the results. The pdf is attached to the pipeline run... (see below...)

PR #485, opened 2025-06-05
[Fix] avoid to encode the realm when generating the authenticate url

SegFault89:
This merge request addresses an issue where the authenticate endpoint URL was incorrectly generated, preventing the use of multi-level or nested realm paths (e.g., /my/pathTo/realm).

Previously, the Uri.Builder.appendPath() method was used for the realm segment. This method URL-encodes characters, converting forward slashes (/) into %2F. As a result, ForgeRock Access Management (AM) would interpret root/MyOrganizationLogin as a single literal realm name, instead of a path to a sub-realm, leading to authentication failures for nested realms.


SegFault89:
For additional context: we've observed that the ForgeRock iOS SDK (at least version 4.6.0) appears to already correctly handle and accept realm paths

PR #520, opened 2026-06-10
SDKS-5120: Add authenticationValidityDuration support to device binding and signing

vibhorgoswami:
# JIRA Ticket
[SDKS-5120](https://pingidentity.atlassian.net/browse/SDKS-5120)

# Description
- Adds authenticationValidityDuration field (default: 5 seconds) to DeviceBindingCallback and DeviceSigningVerifierCallback, giving callers control over how long the generated key remains valid for biometric/device
  authentication.
- Threads the value through to CryptoKey(userId, authenticationValidityDuration) via the DeviceAuthenticator.initialize extension, so the Android Keystore key is generated with the correct user-authentication validity window.
- Adds unit tests to DeviceBindingCallbackTest and DeviceSigningVerifierCallbackTest covering the default value and verifying the custom value propagates to CryptoKey.timeout.

- [x] Coded to standards.
- [x] Ensure backward compatibility.
- [x] API reference docs is created or updated, if applicable.
- [x] Unit tests are written or updated.
- [ ] Integration tests are written, if applicable.

### DIFF
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
@@ -104,6 +104,12 @@ open class DeviceBindingCallback : AbstractCallback, Binding {
     lateinit var attestation: Attestation
         private set
 
+    /**
+     * The duration (in seconds) for which the generated key remains valid for user authentication.
+     * Passed to [CryptoKey] during key generation. Defaults to 5 seconds.
+     */
+    var authenticationValidityDuration: Int = 5
+
     init {
         //If attestation is not provided, default to NONE
         if (!::attestation.isInitialized) {
@@ -222,7 +228,7 @@ open class DeviceBindingCallback : AbstractCallback, Binding {
                                      .build().identifier,
                                  prompt: Prompt? = null) {
 
-        deviceAuthenticator.initialize(userId, prompt ?: Prompt(title, subtitle, description))
+        deviceAuthenticator.initialize(userId, authenticationValidityDuration, prompt ?: Prompt(title, subtitle, description))
 
         if (deviceAuthenticator.isSupported(context, attestation).not()) {
             handleException(DeviceBindingException(Unsupported()))
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 - 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2022 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -9,6 +9,7 @@ package org.forgerock.android.auth.callback
 import android.content.Context
 import android.util.Base64
 import androidx.annotation.Keep
+import org.forgerock.android.auth.DEFAULT_AUTHENTICATION_VALIDITY_DURATION
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
@@ -105,10 +106,24 @@ open class DeviceBindingCallback : AbstractCallback, Binding {
         private set
 
     /**
-     * The duration (in seconds) for which the generated key remains valid for user authentication.
-     * Passed to [CryptoKey] during key generation. Defaults to 5 seconds.
+     * The duration in seconds for which the generated key remains valid for user authentication
+     * without requiring a fresh biometric/credential prompt. Only affects biometric-backed key
+     * types ([DeviceBindingAuthenticationType.BIOMETRIC_ONLY] and
+     * [DeviceBindingAuthenticationType.BIOMETRIC_ALLOW_FALLBACK]); ignored for [DeviceBindingAuthenticationType.NONE]
+     * and [DeviceBindingAuthenticationType.APPLICATION_PIN].
+     *
+     * Must be greater than 0. The Android Keystore treats 0 and -1 as special values that disable
+     * the time-based window and require fresh user authentication on every single key use — which
+     * is very different from any positive duration. Setting this to 0 or -1 will throw
+     * [IllegalArgumentException].
+     *
+     * Defaults to [DEFAULT_AUTHENTICATION_VALIDITY_DURATION] seconds.
      */
-    var authenticationValidityDuration: Int = 5
+    var authenticationValidityDuration: Int = DEFAULT_AUTHENTICATION_VALIDITY_DURATION
+        set(value) {
+            validateAuthenticationValidityDuration(value)
+            field = value
+        }
 
     init {
         //If attestation is not provided, default to NONE
--- a/forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindAuthenticators.kt
+++ b/forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindAuthenticators.kt
@@ -72,7 +72,10 @@ interface DeviceAuthenticator {
      * @param kid Generated kid from the Preference
      * @param userId userId received from server
      * @param challenge challenge received from server
-     * @param customClaims A map of custom claims to be added to the jws payload
+     * @param expiration token expiration time
+     * @param attestation Attestation type to decide whether to include certificate chain in the JWS header
+     *
+     * @return the signed JWT as String
      */
     fun sign(context: Context,
              keyPair: KeyPair,
@@ -89,7 +92,7 @@ interface DeviceAuthenticator {
         if (attestation !is Attestation.None) {
             builder.x509CertChain(getCertificateChain(userId))
         }
-        val jwk = builder.build();
+        val jwk = builder.build()
         val signedJWT = SignedJWT(JWSHeader.Builder(parse(getAlgorithm()))
             .keyID(kid).jwk(jwk).build(),
             JWTClaimsSet.Builder().subject(userId)
@@ -210,7 +213,16 @@ interface DeviceAuthenticator {
 
 }
 
-fun DeviceAuthenticator.initialize(userId: String, prompt: Prompt): DeviceAuthenticator {
+/**
+ * Initialize the DeviceAuthenticator with userId, authentication validity duration and prompt
+ * @param userId The user ID for which the keys will be generated.
+ * @param authenticationValidityDuration The duration (in seconds) for which the generated key remains valid
+ * for user authentication. Passed to the underlying crypto key during key generation. Defaults to 5 seconds.
+ * @param prompt The Prompt to modify the title, subtitle, description
+ *
+ * @return The initialized DeviceAuthenticator instance.
+ */
+fun DeviceAuthenticator.initialize(userId: String, authenticationValidityDuration: Int,  prompt: Prompt): DeviceAuthenticator {
 
     //Inject objects
     if (this is BiometricAuthenticator) {
@@ -219,15 +231,23 @@ fun DeviceAuthenticator.initialize(userId: String, prompt: Prompt): DeviceAuthen
             prompt.description,
             deviceBindAuthenticationType = this.type()))
     }
-    initialize(userId)
+    initialize(userId, authenticationValidityDuration)
     this.prompt(prompt)
     return this
 }
 
-fun DeviceAuthenticator.initialize(userId: String): DeviceAuthenticator {
+/**
+ * Initialize the DeviceAuthenticator with userId and authentication validity duration
+ * @param userId The user ID for which the keys will be generated.
+ * @param authenticationValidityDuration The duration (in seconds) for which the generated key remains valid
+ * for user authentication. Passed to the underlying crypto key during key generation. Defaults to 5 seconds.
+ *
+ * @return The initialized DeviceAuthenticator instance.
+ */
+fun DeviceAuthenticator.initialize(userId: String, authenticationValidityDuration: Int = 5): DeviceAuthenticator {
     //Inject objects
     if (this is CryptoAware) {
-        this.setKey(CryptoKey(userId))
+        this.setKey(CryptoKey(userId, authenticationValidityDuration))
     }
     return this
 }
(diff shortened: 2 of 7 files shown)


forgerock-core/src/main/java/org/forgerock/android/auth/CryptoKey.kt
```diff
@@ -25,12 +25,11 @@ import java.security.spec.AlgorithmParameterSpec
 /**
  * Helper class to generate and sign the keys
  */
-class CryptoKey(private var keyId: String) {
+class CryptoKey(private var keyId: String, val timeout: Int = 5) {
```
george-bafaloukas-forgerock, line 31: Worth calling out that the Android Keystore treats `timeout` values of `0` and `-1` with distinct semantics: both cause the key to require fresh biometric authentication on *every single use* (no time-based window), which is very different from any positive value. Now that this is developer-configurable, it might be worth adding a `require(timeout > 0) { "authenticationValidityDuration must be a positive integer" }` guard to surface misuse early with a clear message, or at minimum documenting the special-value behaviour in the KDoc so integrators aren't caught off guard. The customer driving this change is likely to set values like `30` or `60`, but future callers might not realise what `0` does.

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindAuthenticators.kt
```diff
@@ -210,7 +213,16 @@ interface DeviceAuthenticator {
 
 }
 
-fun DeviceAuthenticator.initialize(userId: String, prompt: Prompt): DeviceAuthenticator {
+/**
+ * Initialize the DeviceAuthenticator with userId, authentication validity duration and prompt
+ * @param userId The user ID for which the keys will be generated.
+ * @param authenticationValidityDuration The duration (in seconds) for which the generated key remains valid
+ * for user authentication. Passed to the underlying crypto key during key generation. Defaults to 5 seconds.
+ * @param prompt The Prompt to modify the title, subtitle, description
```
george-bafaloukas-forgerock, line 222: Small doc nit: this `@param` says "Defaults to 5 seconds" but `authenticationValidityDuration` is a *required* parameter in this three-argument overload — it has no default value. The `= 5` default only lives on the two-argument overload below. Worth removing that phrase here to avoid misleading callers.

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindAuthenticators.kt
```diff
@@ -210,7 +213,16 @@ interface DeviceAuthenticator {
...
+ *
+ * @return The initialized DeviceAuthenticator instance.
+ */
+fun DeviceAuthenticator.initialize(userId: String, authenticationValidityDuration: Int,  prompt: Prompt): DeviceAuthenticator {
```
george-bafaloukas-forgerock, line 227: Looks like there's a stray double-space before `prompt` in this signature: `authenticationValidityDuration: Int,  prompt: Prompt`. Easy cosmetic fix!

george-bafaloukas-forgerock (commented):
Great work on this — the core implementation is clean and the propagation through `initialize()` → `CryptoKey` is well structured. A couple of small things worth addressing before merging:

**`DeviceSigningVerifierCallbackTest.kt` — missing `@Test` annotations**: Two pre-existing test methods in this file — `testSignForForValidClaims()` and `testSignForForInvalidClaims()` — are missing `@Test` annotations, so JUnit silently skips them entirely. Since this PR is already touching the file, it would be a great time to fix those here. The `testSignForForInvalidClaims` case is especially worth addressing as it covers the `InvalidCustomClaims` error path.

witrisna (dismissed):

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindAuthenticators.kt
vibhorgoswami: Updated and removed the space.

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindAuthenticators.kt
vibhorgoswami: Updated docs

forgerock-core/src/main/java/org/forgerock/android/auth/CryptoKey.kt
vibhorgoswami: The change has now been implemented in [DeviceSigningVerifierCallback.kt‎](https://github.com/ForgeRock/forgerock-android-sdk/pull/520/changes#diff-3a4e37d13f04f0deaff14daf19f5b1055ce5baedf6486689da04267ce3ac5ccc)
```kotlin
var authenticationValidityDuration: Int = 5
        set(value) {
            require(value > 0) { "authenticationValidityDuration must be greater than 0" }
            field = value
        }
```

george-bafaloukas-forgerock (commented):
## Re-review (latest commit: guard authentication validity duration)

The setter guard + tests are a solid addition. The core threading (`callback → initialize(…, duration, …) → CryptoKey(userId, timeout)`) is correct, the guard correctly rejects the dangerous `0`/`-1` Keystore special values, and `timeout` (server-sent, used for `withTimeout`/JWT exp) is kept properly distinct from `authenticationValidityDuration`. A few items worth a look:

### 1. Binary-incompatible change to a published class — contradicts the "backward compatibility" DoD
`forgerock-core/.../CryptoKey.kt` — `CryptoKey` is a public class in the published `forgerock-core` artifact. Changing the primary constructor from `CryptoKey(keyId: String)` to `CryptoKey(keyId: String, timeout: Int = 5)` changes its JVM signature from `<init>(String)` to `<init>(String, int)`. Kotlin default parameters do **not** generate a synthetic overload — there's no `@JvmOverloads`, and there's no binary-compatibility validator in the build. Any pre-compiled third-party code calling `new CryptoKey(userId)` would hit `NoSuchMethodError` at runtime until recompiled. Since this PR ticks "Ensure backward compatibility," could we make a conscious call here — add `@JvmOverloads`, or confirm `CryptoKey` is not treated as consumer-facing API? (Intra-SDK callers recompile, so no internal break.)

### 2. `authenticationValidityDuration` is a no-op for `NONE` and `APPLICATION_PIN`, and the test masks it
Only the biometric authenticators consume `cryptoKey.timeout` (via `setUserAuthenticationParameters` / `setUserAuthenticationValidityDurationSeconds` in `BiometricOnly` / `BiometricAndDeviceCredential`). `None` and `ApplicationPinDeviceAuthenticator` never read it. That's defensible, but `testAuthenticationValidityDurationCustomValuePassedToAuthenticator` asserts the value on a `mock<None>()` — i.e. it verifies plumbing into `CryptoKey` while using the one authenticator type that ignores the value. Suggest documenting in the public KDoc that the setting only affects biometric-backed keys, and/or covering a biometric type in the test.

### 3. Property + validation + KDoc duplicated verbatim across both callbacks
`DeviceBindingCallback.kt` and `DeviceSigningVerifierCallback.kt` — the `var authenticationValidityDuration` + `require(value > 0)` setter + KDoc are copy-pasted into both classes, which both implement `Binding`. A Kotlin interface can't hold a backing field, but the validation could live in a shared helper (e.g. `Binding.validateValidityDuration(Int)`) to keep the two copies from drifting (e.g. one becoming `>= 0`).

### 4. Default `5` duplicated in three places
The magic default appears in `CryptoKey(timeout: Int = 5)`, `initialize(userId, authenticationValidityDuration: Int = 5)`, and both callbacks' `= 5`. A single shared constant would keep them from diverging.

---

No blocking correctness bug — #1 (binary compatibility) is the one I'd want an explicit decision on before merge given the DoD claim; the rest are advisory. (Nice catch incidentally adding `@Test` to the two previously-unannotated `testSignFor*` methods so they actually run now.)

forgerock-core/src/main/java/org/forgerock/android/auth/CryptoKey.kt
```diff
@@ -25,12 +25,11 @@ import java.security.spec.AlgorithmParameterSpec
...
```
george-bafaloukas-forgerock: **Binary-incompatible change to a published class — contradicts the "backward compatibility" DoD.**

`CryptoKey` is a public class in the published `forgerock-core` artifact. Changing the primary constructor from `CryptoKey(keyId: String)` to `CryptoKey(keyId: String, timeout: Int = 5)` changes its JVM signature from `<init>(String)` to `<init>(String, int)`. Kotlin default parameters do **not** generate a synthetic overload — there's no `@JvmOverloads` here and no binary-compatibility validator in the build. Any pre-compiled third-party code calling `new CryptoKey(userId)` would hit `NoSuchMethodError` at runtime until recompiled.

Since this PR ticks "Ensure backward compatibility," could we make a conscious call here — add `@JvmOverloads`, or confirm `CryptoKey` is not treated as consumer-facing API? (Intra-SDK callers recompile, so no internal break.)

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindAuthenticators.kt
```diff
@@ -219,15 +231,23 @@ fun DeviceAuthenticator.initialize(userId: String, prompt: Prompt): DeviceAuthen
...
+ * @param authenticationValidityDuration The duration (in seconds) for which the generated key remains valid
+ * for user authentication. Passed to the underlying crypto key during key generation. Defaults to 5 seconds.
+ *
+ * @return The initialized DeviceAuthenticator instance.
+ */
+fun DeviceAuthenticator.initialize(userId: String, authenticationValidityDuration: Int = 5): DeviceAuthenticator {
     //Inject objects
     if (this is CryptoAware) {
-        this.setKey(CryptoKey(userId))
+        this.setKey(CryptoKey(userId, authenticationValidityDuration))
```
george-bafaloukas-forgerock, line 254: **`authenticationValidityDuration` is a no-op for `NONE` and `APPLICATION_PIN`.**

Only the biometric authenticators consume `cryptoKey.timeout` (via `setUserAuthenticationParameters` / `setUserAuthenticationValidityDurationSeconds` in `BiometricOnly` / `BiometricAndDeviceCredential`). `None` and `ApplicationPinDeviceAuthenticator` never read it. That's defensible, but `testAuthenticationValidityDurationCustomValuePassedToAuthenticator` asserts the value on a `mock<None>()` — i.e. it verifies plumbing into `CryptoKey` while using the one authenticator type that ignores the value, giving false confidence the feature "works for NONE." Suggest documenting in the public KDoc that the setting only affects biometric-backed keys, and/or covering a biometric type in the test.

forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
```diff
@@ -104,6 +104,23 @@ open class DeviceBindingCallback : AbstractCallback, Binding {
...
+     * without requiring a fresh biometric/credential prompt.
+     *
+     * Must be greater than 0. The Android Keystore treats 0 and -1 as special values that disable
+     * the time-based window and require fresh user authentication on every single key use — which
+     * is very different from any positive duration. Setting this to 0 or -1 will throw
+     * [IllegalArgumentException].
+     *
+     * Defaults to 5 seconds.
+     */
+    var authenticationValidityDuration: Int = 5
```
george-bafaloukas-forgerock: **Duplicated property + validation + KDoc across both callbacks.**

This `var authenticationValidityDuration` + `require(value > 0)` setter + KDoc block is copy-pasted verbatim into `DeviceSigningVerifierCallback` too. Both classes implement `Binding`. A Kotlin interface can't hold a backing field, but the validation could live in a shared helper (e.g. `Binding.validateValidityDuration(Int)`) so the two copies can't drift (e.g. one becoming `>= 0`).

forgerock-core/src/main/java/org/forgerock/android/auth/CryptoKey.kt
george-bafaloukas-forgerock: **Default `5` duplicated in three places.** This magic default also appears in `initialize(userId, authenticationValidityDuration: Int = 5)` and both callbacks' `= 5`. A single shared constant would keep them from diverging.

george-bafaloukas-forgerock (commented):
Re-review of the latest changes — the setter guard + tests are a solid addition and the core threading is correct. Four inline notes below; #1 (binary compatibility) is the only one I'd want an explicit decision on before merge given the DoD claim.

forgerock-core/src/main/java/org/forgerock/android/auth/CryptoKey.kt
vibhorgoswami: Updated to use an internal constant.

forgerock-auth/src/main/java/org/forgerock/android/auth/callback/DeviceBindingCallback.kt
vibhorgoswami: Addressed

forgerock-auth/src/main/java/org/forgerock/android/auth/devicebind/DeviceBindAuthenticators.kt
vibhorgoswami: 1. KDoc updated on both callbacks — the authenticationValidityDuration doc now explicitly states:
  Only affects biometric-backed key types (BIOMETRIC_ONLY and BIOMETRIC_ALLOW_FALLBACK); ignored for NONE and APPLICATION_PIN.
  2. Tests switched to BiometricOnly — both testAuthenticationValidityDurationCustomValuePassedToAuthenticator tests (in DeviceBindingCallbackTest and DeviceSigningVerifierCallbackTest) now mock BiometricOnly instead of None,
  so they verify the value flows through to CryptoKey.timeout on the authenticator type that actually consumes it.

forgerock-core/src/main/java/org/forgerock/android/auth/CryptoKey.kt
vibhorgoswami: Added JvmOverloads to add additional java constructor.

forgerock-core/src/main/java/org/forgerock/android/auth/CryptoKey.kt
vibhorgoswami: Updated to use `@JvmOverloads` and also added a constant to be a single source of truth.

george-bafaloukas-forgerock (approved):
LGTM

spetrov (approved):
Changes look good to me! 👍🏻

PR #521, opened 2026-06-12
[fix] SDKS-5114 Handle AM 400 for Push Number Challenge with distinct exception

rodrigoareis:
# JIRA Ticket

[SDKS-5114](https://pingidentity.atlassian.net/browse/SDKS-5114) "[Legacy][Android] forgerock-authenticator handle AM 400 for Push Number Challenge"

# Description

AM 8.1.0 added server-side validation for Push Number Challenge: selecting the wrong number now yields `400 Bad Request` on the authenticate endpoint instead of `200`. Previously `PushResponder.performAuthentication(...)` treated any non-200 as a generic `PushMechanismException("Communication with server returned <code> code.")`, so users got no semantic signal and the app could not distinguish a wrong-number rejection from a network failure. 

This change introduces a new `PushNumberChallengeException extends PushMechanismException` subtype and wires it into the `onResponse` callback: when the response code is `400` **and** the notification is `PushType.CHALLENGE`, throw the distinct exception and leave the notification `pending = true` / `approved = false` so the app can retry. All other code paths (200 success, non-CHALLENGE 400, 401/500/etc.) keep their current behavior exactly.

- [x] Coded to standards.
- [ ] API reference docs is created or updated, if applicable.
- [x] Unit tests are written or updated.
- [ ] Integration tests are written, if applicable.

### DIFF
--- a/forgerock-authenticator/src/main/java/org/forgerock/android/auth/PushResponder.java
+++ b/forgerock-authenticator/src/main/java/org/forgerock/android/auth/PushResponder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020 - 2025 Ping Identity Corporation. All rights reserved.
+ * Copyright (c) 2020 - 2026 Ping Identity Corporation. All rights reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
@@ -23,6 +23,7 @@ import com.nimbusds.jwt.SignedJWT;
 
 import org.forgerock.android.auth.exception.ChallengeResponseException;
 import org.forgerock.android.auth.exception.PushMechanismException;
+import org.forgerock.android.auth.exception.PushNumberChallengeException;
 import org.forgerock.android.auth.util.RequestBuilder;
 import org.jetbrains.annotations.NotNull;
 import org.json.JSONException;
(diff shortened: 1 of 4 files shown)


vibhorgoswami (dismissed):
LGTM

forgerock-authenticator/src/main/java/org/forgerock/android/auth/PushResponder.java
```diff
@@ -354,6 +355,10 @@ public void onResponse(@NotNull Call call, @NotNull Response response) {
                         return;
                     }
                     listener.onSuccess(null);
+                } else if (response.code() == 400 && pushNotification.getPushType() == PushType.CHALLENGE) {
+                    // Number challenge failed — notification stays pending so the user can retry.
+                    listener.onException(new PushNumberChallengeException(
```
spetrov: Isn't this hardcoded message `Number challenge failed: the selected number was incorrect.` too specific?... The 400 response on a CHALLENGE notification could be for other reasons (expired challenge or malformed JWT payload for example...)

In fact, maybe it would be better to directly surface the message coming from AM - i.e. with the fix AM now returns this: 
```
{
  "code": 400,
  "reason": "Bad Request",
  "message": "Number challenge predicate not met."
}
```
Also, consider adding a WARN level message in the logs...?

witrisna (dismissed):

spetrov (dismissed):
LGTM. Just one question to clarify...

spetrov (approved):
LGTM! 👍🏻

